PR #23, opened 2015-01-23
Go vetting

jszwedko:
This PR fixes issues `go vet` brings up.

I also added `go vet` to Travis.

@lionelbarrow you may want to consider generating new API keys and encrypting them in the `travis.yml`. Not sure how much it matters for sandbox accounts.


lionelbarrow:
I don't think it matters a lot for sandbox accounts. I also like that outside contributors can use the pre-existing account if necessary -- until someone does something bad, I'd prefer to trust the community.

PR #26, opened 2015-01-23
Use real time.Time structs for CreatedAt/UpdatedAt

jszwedko:
Also introduce Date struct to handle values from Braintree that are
dates rather than datetimes.

This breaks the API, but makes these fields much more useful to the end-user (otherwise they have to deserialize the result further).

Thoughts on this breaking change?


lionelbarrow:
Lets do this and the `big.Rat` change in the same major version bump.

lionelbarrow:
I moved the Date struct to its own package.

Issue #27, opened 2015-01-23
Use a "nullable" type for primitive values that may be empty in Braintree responses


jszwedko:
E.g. `Plan.NumberOfBillingCycles` may be empty in the response XML from Braintree, but is otherwise an integer. We currently deserialize to a `string` to avoid this issue.

Possible solutions:
- Create types that can either be an `int` (or `bool`, `float`, etc.) or `nil` and use those (similar to `sql.NullInt`)
- Make use of https://github.com/guregu/null which does this work, but would require us to introduce a third-party dependency and expose that dependency to the end user

jszwedko:
I'm currently leaning towards rolling our own to avoid the dependency.

lionelbarrow:
Done.

PR #28, opened 2015-01-24
Add Null(Int64|Float64|Bool) types to support empty XML elements

jszwedko:
These fields can be used with elements that either have a primitive
value (i.e. float64, int64, bool), but where they can be empty in the
response from Braintree.

This is a possible way to address #27 

/cc @lionelbarrow

### DIFF
--- a/disbursement_details.go
+++ b/disbursement_details.go
@@ -5,11 +5,11 @@ import (
 )
 
 type DisbursementDetails struct {
-	XMLName                        xml.Name `xml:"disbursement-details"`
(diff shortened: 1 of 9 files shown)


jszwedko:
This an API breaking change, but I think it offers a much better interface to the user than asking them to do the deserialization of the fields on their end.

jszwedko:
We could move this into a subpackage first if we decide to go with this implementation.

lionelbarrow:
It seems like Decimals are preferable to NullFloats in a lot of these amount fields?

PR #30, opened 2015-02-06
Use fixed precision data type for amounts

jszwedko:
To avoid the loss of precision that might occur from using floats.


lionelbarrow:
Awesome. I'll get started on converting the other float types to decimals.

Issue #24, opened 2015-01-23
Use `big.Rat` or another type to avoid losing precision for currency values


jszwedko:
I know this would break the API, but how do you feel about swapping out the `float64`s currently being used for amounts with `big.Rat`s to avoid losing precision?

lionelbarrow:
It's something I've been meaning to get around to. This was an oversight when I first wrote the library.

lionelbarrow:
Done!

PR #31, opened 2015-02-20
More decimals

jszwedko:
Replace more decimals with concreted Decimal type rather than float


jszwedko:
Addresses #24

jszwedko:
I think that above error is transient (possibly due to running multiple integration tests in parallel with the same merchant account?).

lionelbarrow:
Merged it in. Now for the Nullable types.

PR #36, opened 2015-04-28
Implement customer search

pranavraja:
See https://www.braintreepayments.com/docs/ruby/customers/search


lionelbarrow:
Awesome, thank you!

PR #40, opened 2015-07-23
add refund point and refund tests

codewitch:
Adding refund capabilities and tests to prove it works :]


lionelbarrow:
:+1: Thanks!

PR #41, opened 2015-07-30
Tj refund ids

codewitch:
Adding refund ids to the transaction return object.

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -6,26 +6,28 @@ import (
 )
 
 type Transaction struct {
-	XMLName             string               `xml:"transaction"`
-	Id                  string               `xml:"id,omitempty"`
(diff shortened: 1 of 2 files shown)


lionelbarrow:
You've got a typo on the struct tag in refunded transaction ID - `refundied`. I think you should add some tests before I merge this. It should be really easy to extend the test `TestTransactionCreateSettleAndFullRefund` with assertions that cover these fields.

lionelbarrow:
Sorry, didn't mean to hit the close button :) Just say something in the thread when you've added some tests.

codewitch:
Thanks for the suggestion! I fixed the typo and made checks in the tests. Let me know if you have any other comments :)

lionelbarrow:
Awesome, thanks!

PR #43, opened 2015-08-08
Adding support for CustomerId field on client token requests (to support vault workflow)

dbudworth:
Simple change that adds CustomerID field to the ClientToken struct
Also renamed ClientToken struct to ClientTokenRequest as that more accurately describes it.


lionelbarrow:
Looks good! I'm going to slightly tweak the API and add another test.

PR #44, opened 2015-08-10
Add support for paypal payment method

maxsz:


lionelbarrow:
This looks great! I'm going to tweak the tests a bit though -- I don't think it's correct to simply not run the tests unless the right environmental variable is set. We can get around the environmental variable stuff with some special nonce values as in this test: https://github.com/braintree/braintree_ruby/blob/master/spec/integration/braintree/paypal_account_spec.rb#L6-L28

maxsz:
Hi, thanks for the feedback. We are already using this for some time now. I also had to make some other changes to get the braintree-go library to work for us in production. Since I didn't here anything back from you for this pull request I didn't make any other pull requests for these changes. Would you be interested in this? Also, are you considering to put some more work into this and make it an official library?

olivoil:
@maxsz @lionelbarrow maybe this would be a good time to add an interface for payment methods, something like:

```
type PaymentMethod interface {
	GetToken() string
	IsDefault() bool
	GetImageURL() string
	GetCustomerID() string
}
```

and add a generic `PaymentMethodGateway` that can be used without knowing if it's a credit card or paypal?

lionelbarrow:
@maxsz I don't have any timeline or word on official support for Go from Braintree. I wrote this library for fun but haven't had as much time to maintain it as I once did. That said, I'd be very interested in any contributions you can send back.

vially:
Any chance to get this pull request merged? I have done some work on implementing a generic [payment method gateway](https://github.com/vially/braintree-go/tree/payment-method) and I would like to propose it for a pull request but it depends on this PR being merged first.

otraore:
No updates on this yet?

dannyvankooten:
Sorry for bumping but it would be really neat if this could get merged @lionelbarrow. Reckon you are super busy, perhaps granting @vially or someone else commit rights is a viable option?

maxsz:
Maybe if @lionelbarrow doesn't have the time, @leighmcculloch could help us out here? Otherwise it might make sense to start a new maintained go-to fork @vially ?

vially:
Given the lack of responses on this and most other pull requests I think starting a new fork is a valid option. If/when this repo starts being updated again we can reconsider the options. Until them I think a new maintained go-to fork is the recommended path forward.

leighmcculloch:
Thanks for putting together this PR @maxsz. It looks great and just has one issue with needing a different api version to successfully find PayPal accounts and we wanted to tweak the tests so that they can run against any Braintree sandbox account without needing to pre-create the PayPal Account payment method manually.

Because you've been waiting a while for a response I want to reduce the back and forth and I've gone ahead and made the changes needed to get it merged while preserving your commit in the history. We should be able to merge #96 soon which will automatically mark this PR as merged. Comments appreciated on #96.

PR #47, opened 2015-10-07
Added ReassignId function in customer_gateway.go to allow Customer Id changes

AlasdairF:


AlasdairF:
There are 2 new functions in these commits. The first is ReassignId for changing a customerId. The second is much more useful: a Verify function for generating to correct response to the webhook bt_challenge.

lionelbarrow:
I don't think we necessarily want the `ReassignId` function -- it seems like a subset of the update functionality that already exists. I'll merge in the webhook verification stuff though.

AlasdairF:
Hi Lionel, currently it is not possible to update the customer ID using your `func (g *CustomerGateway) Update(c *Customer) (*Customer, error)` because it uses the customer ID from the *Customer object and so it can never change the customer's ID. This is why I created the ReassignId function as it is the only way to actually change a customer's ID.

lionelbarrow:
Ah, that's pretty annoying. Maybe the update method should take the current record ID as an argument?

AlasdairF:
Yes, that would be a better solution. The only problem is that it would force everyone using this package to update their implementation.

It is probably best though: to change the Update function to use a customer ID parameter.

PR #52, opened 2015-11-04
add support for oauth gateway

olivoil:
I think a first pass at it could look something like this.

What do you think @lionelbarrow ?

### DIFF
--- a/braintree.go
+++ b/braintree.go
@@ -40,12 +40,65 @@ func New(env Environment, merchId, pubKey, privKey string) *Braintree {
 }
 
 type Braintree struct {
-	Environment Environment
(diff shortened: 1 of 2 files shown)


olivoil:
fixes #52

olivoil:
Braintree told me the oauth/partners api is closed to new registrations at the moment, and to use the marketplace api instead. So my interest in supporting it is limited. Let me know if there is an interest to finish this, otherwise feel free to close the PR.

lionelbarrow:
Yea, I agree -- I don't think this is something we'd want to maintain given that nobody will use it for a while. I'm gonna close it. Thanks for doing the work though!

PR #53, opened 2015-11-10
Marshal scaled zero decimal correctly

dhamidi:
Calling MarshalText on `NewDecimal(0, 2)` would panic because of an
out-of-bounds slice access.  The expected result of marshaling this
decimal is "0.00".

This commit checks the length of the string representing the unscaled
number and pads it with zero if the string is shorter than the scale.


dhamidi:
I ran into this error after trying to dump a received webhook notification from Braintree's sandbox (for the subscription created event) as JSON.

lionelbarrow:
Great catch. Thank you!

PR #56, opened 2016-02-03
Customizable BaseURL when in Development Environment

cjrd:
This small PR makes the BaseURL customizable when the Braintree Environment is `Development`. This small change allows this library to play nicely with [httptest](https://golang.org/pkg/net/http/httptest/) when testing various HTTP responses.


lionelbarrow:
Hmm, I think the better thing to do here is to change the modeling of how environments work rather than force this in. (Looking at the way I originally laid environments out, I think I made it too brittle.) How about something like this?

```go
type Environment string

const (
  Development Environment = "http://localhost:3000"
  Sandbox Environment = "https://sandbox.braintreegateway.com"
  Production Environment = "https://www.braintreegateway.com"
)

...

func (g *Braintree) BaseURL() string {
  return g.environment
}
```

Then you can customize the URL you use by making a new instance of `Environment`.

cjrd:
I agree - I think this is a better solution.

PR #60, opened 2016-03-17
Fixed decimal bug where NewDecimal(5,2) would print 5.00 instead of .05

knifeedge:


lionelbarrow:
Wow, that's nasty. Thanks very much for the patch.

PR #61, opened 2016-04-01
Added go1.5 and go1.6 to test runs.

leighmcculloch:
What
===
Added go1.5 and go1.6 to test runs.

Why
===
So that we identify any issues on go1.5 and go1.6 ahead of merging PRs.


lionelbarrow:
Yea

PR #63, opened 2016-04-13
Add more XML Transaction fields

fabriziomoscon:
- additional-processor-response
- risk-data
- payment-instrument-type

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -6,33 +6,36 @@ import (
 )
 
 type Transaction struct {
-	XMLName                    string               `xml:"transaction"`
-	Id                         string               `xml:"id,omitempty"`
-	CustomerID                 string               `xml:"customer-id,omitempty"`
-	Status                     string               `xml:"status,omitempty"`
-	Type                       string               `xml:"type,omitempty"`
-	Amount                     *Decimal             `xml:"amount"`
-	OrderId                    string               `xml:"order-id,omitempty"`
-	PaymentMethodToken         string               `xml:"payment-method-token,omitempty"`
-	PaymentMethodNonce         string               `xml:"payment-method-nonce,omitempty"`
-	MerchantAccountId          string               `xml:"merchant-account-id,omitempty"`
-	PlanId                     string               `xml:"plan-id,omitempty"`
-	CreditCard                 *CreditCard          `xml:"credit-card,omitempty"`
-	Customer                   *Customer            `xml:"customer,omitempty"`
-	BillingAddress             *Address             `xml:"billing,omitempty"`
-	ShippingAddress            *Address             `xml:"shipping,omitempty"`
-	Options                    *TransactionOptions  `xml:"options,omitempty"`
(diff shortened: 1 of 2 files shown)


lionelbarrow:
Hey @fabriziomoscon,

Thanks very much for the patch! For these features, can you add a either or an integration test that uses these features successfully or a unit test that shows this code deserializes the Braintree XML? You can see an example of the latter in the [error parsing test](https://github.com/lionelbarrow/braintree-go/blob/master/errors_test.go).

fabriziomoscon:
Hi @lionelbarrow,

I have added few more check in the file `transaction_integration_test.go` for:
- RiskData
- AdditionalProcessorResponse

I was going to add some test for `PaymentInstrumentType` in case of `ApplePay` but the sandbox seems to return empty in all cases. Could you advise what is the best way to check this unmarshal and most importantly if I am right checking this xml field to distinguish ApplePay transactions.

transaction.go
```diff
@@ -6,33 +6,36 @@ import (
...
+	DisbursementDetails         *DisbursementDetails `xml:"disbursement-details,omitempty"`
+	RefundId                    string               `xml:"refund-id,omitempty"`
+	RefundIds                   *[]string            `xml:"refund-ids>item,omitempty"`
+	RefundedTransactionId       *string              `xml:"refunded-transaction-id,omitempty"`
+	ProcessorResponseCode       int                  `xml:"processor-response-code,omitempty"`
+	ProcessorResponseText       string               `xml:"processor-response-text,omitempty"`
+	AdditionalProcessorResponse string               `xml:"additional-processor-response,omitempty"`
+	ProcessorAuthorizationCode  string               `xml:"processor-authorization-code,omitempty"`
+	SettlementBatchId           string               `xml:"settlement-batch-id,omitempty"`
+	PaymentInstrumentType       string               `xml:"paymentInstrumentType,omitempty"`
```
leighmcculloch: `paymentInstrumentType` should be `payment-instrument-type`.

transaction_integration_test.go
```diff
@@ -245,6 +252,18 @@ func TestAllTransactionFields(t *testing.T) {
...
+	}
+	if tx2.RiskData == nil {
+		t.Fatal("expected tx2.RiskData not to be empty")
+	}
+	if tx2.RiskData.ID == "" {
+		t.Fatal("expected tx2.RiskData.ID not to be empty")
+	}
+	if tx2.RiskData.Decision != "Approve" {
+		t.Fatalf("expected tx2.RiskData.Decision to be Approve, but got %s", tx2.RiskData.Decision)
+	}
```
leighmcculloch: Could you add an expectation in here for `tx2.PaymentInstrumentType` to equal `credit_card`? This test checks all the fields on `Transaction`, but `PaymentInstrumentType` is missing here.

leighmcculloch:
@fabriziomoscon: Actually if you rebase/merge master into this you'll get the payment-instrument-type because that already exists in master now.

PR #64, opened 2016-04-15
Fix typos in nullable type tests.

leighmcculloch:
What
===
Fix `sholud` typos in nullable type tests that should be `should`.

Why
===
Clean, clear code.


lionelbarrow:
👍

PR #65, opened 2016-04-15
Fix tests that depend on math/rand for uniqueness. 

leighmcculloch:
What
===
Fix tests that depend on math/rand for uniqueness by replacing the unix timestamp based random number generation with crypto random.

Why
===
The tests were using math/rand to make the test data that is created in the sandbox environment unique to that test run. Tests were failing often on TravisCI because the tests are run on multiple versions of Go at the same time. At roughtly the same time the test was running, and creating transactions that had the same customer name because the tests start at the same time resulting in the unix timestamp being the same. Using crypto/rand significantly decreases the possibility of collision.

### DIFF
--- a/merchant_account_integration_test.go
+++ b/merchant_account_integration_test.go
@@ -2,21 +2,19 @@ package braintree
 
 import (
 	"encoding/xml"
-	"math/rand"
-	"strconv"
 	"testing"
-	"time"
+
+	"github.com/lionelbarrow/braintree-go/testhelpers"
 )
 
-var acctId int
+var acctId string
 
 func TestMerchantAccountCreate(t *testing.T) {
-	rand.Seed(time.Now().UTC().UnixNano())
-	acctId = rand.Int() + 1
+	acctId = testhelpers.RandomString()
 	acct := MerchantAccount{
 		MasterMerchantAccountId: testMerchantAccountId,
 		TOSAccepted:             true,
-		Id:                      strconv.Itoa(acctId),
+		Id:                      acctId,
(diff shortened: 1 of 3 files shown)


lionelbarrow:
It seems crazy to me that this was the root cause of the problem. Wouldn't the different simultaneous builds have to start at like, _exactly_ the same time for this to cause problems?

leighmcculloch:
Not exactly. Unix time is in seconds, and a second is a relatively long time for fast go tests. It's also possible that there was some other reason that collision was happening where the data for two tests were getting returned by the API, but I didn't find it and making the inputs truly random should help.

I want to put the `randomUniqueString` function into a shared 'test helper' location. Where's the best place in go tests to put it? Because once I fix it in other locations it'll be used in multiple test files.

lionelbarrow:
Ah, for whatever reason I was under the impression we were getting subsecond precision in the time-based seed. That makes more sense now.

A test helper file sounds like a good idea. IIRC Go has some rules about automatically excluding files ending `_test.go` from non-test build artifacts (see the bit about `TestGoFiles` in  https://golang.org/cmd/go/#hdr-List_packages) so maybe a `helpers_test.go`? I'm not sure if that's worth the confusion though if the functions are private.

leighmcculloch:
@lionelbarrow I placed the test helpers inside a sub-package `testhelpers`. I saw this pattern elsewhere, it's clean and keeps it separate. It isn't imported when building the library itself. Let me know what you think.

lionelbarrow:
Merged. Let's see if the TravisCI problem with the sub-packages persists.

Issue #59, opened 2016-03-16
Escrow support


atishpatel:
Hello! 

I'm just getting started with Braintree for my application, and it's really awesome to see this library that supports Golang for Braintree. However, I need escrow support for my application. 

Is there anyone working on this?

PR #67, opened 2016-04-27
Escrow Support

atishpatel:
This request is to add support for escrow. #59  I've added the functions required by looking at this library and BrainTree's official libraries. 

I'm not extremely familiar with the library so please help. It seems it's not possible to test ReleaseFromEscrow or CancleRelease in a normal environment because I can't create escrow Transactions at will. However, I did write the test based on the other official libraries and comment them out. It'll should work if "transactions/"+id+"/escrow" URL works.

Thank you.


lionelbarrow:
Thanks! This looks very well done, I'll review it now.

lionelbarrow:
@atishpatel, when trying to merge this I ran into an issue in the Braintree API where if the `hold-in-escrow` option flag is passed at all -- even if it's `false` -- the API will reject the request unless the transaction _can_ be held in escrow.

I wrote a patch for the API and it should be deployed relatively soon, but until then I'm going to hold off on merging this.

atishpatel:
@lionelbarrow Looks like i misspelled 'omitempty' to 'omniempty'(the all encompassing empty. haha) for the hold in escrow. So, now it shouldn't be included, but waiting until the patch comes seems like a safe bet.

Thank you! Let me know if there is anything I can do to help.

PR #69, opened 2016-05-13
Add an explicit go_import_path.

leighmcculloch:
What
===
Add an explicit go_import_path pointing to the main repository.

Why
===
When TravisCI runs tests in forks of this package it needs to setup the
environment with the package downloaded to the src path that matches the
main import path, otherwise the absolute imports within the package will
be referencing the non-forked version and be downloaded.

Examples
===
Before
---
screen shot 2016-05-13 at 1 20 29 pm

After
---
screen shot 2016-05-13 at 1 22 01 pm


lionelbarrow:
👍

PR #70, opened 2016-05-13
Fix test that could not be run in parallel.

leighmcculloch:
What
===
Change the Customer test to use a unique value when creating value on the sandbox account that the test then searches for.

Why
===
The test was updating a Customer to have a first name of "John" and searching for all customers with that name. If this test is run in parallel with itself on the same sandbox account it will get results from other runs of this test and fail because the Customer Ids do not match.

### DIFF
--- a/customer_integration_test.go
+++ b/customer_integration_test.go
@@ -2,6 +2,8 @@ package braintree
 
 import (
 	"testing"
+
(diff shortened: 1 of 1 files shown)


lionelbarrow:
👍

PR #75, opened 2016-06-04
Fix parse signature and status code issue for submerchant update

atishpatel:
Braintree webhooks for notifications sometimes provide multiple payload and signature. This adds support for that. 
Also, braintree returns status 200 for submerchant update. I guess they use to return status 201. 

I suggest all the http statuscode checks be converted to check for anything between 200 and 299 or be changed to check the actual body.


leighmcculloch:
@atishpatel: 👋 I'd like to see if I can get the signature fix merged in, but this PR has a lot of other changes. Are you able to whittle it down to just the signature fix?

PR #76, opened 2016-06-22
Add DeviceData to transaction model

cgcote:
https://developers.braintreepayments.com/guides/paypal/server-side/ruby#using-device-data

"Collecting device data from your customers is required when initiating non-recurring transactions from Vault records. Collecting and passing this data with transactions will help reduce decline rates."

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -1,8 +1,9 @@
 package braintree
 
 import (
-	"github.com/lionelbarrow/braintree-go/nullable"
 	"time"
+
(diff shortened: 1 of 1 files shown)


lionelbarrow:
Hey @cgcote, can you add some use of device data to an integration test? Otherwise this looks pretty helpful, thanks!

leighmcculloch:
@cgcote @lionelbarrow This PR hasn't had any activity for the last couple months, so I've picked up the change, added use of the data to the integration test and put it in #80.

PR #78, opened 2016-08-10
Add PaymentInstrumentType field to Transaction struct

brmasson:
The commit adds a field to the Transaction struct.
The test demonstrates how the payment method can be correctly identified in a transaction were the payment method is specified through a nonce.


brmasson:
It seems that the PaymentInstrumentType field is being sent to Braintree despite the 'omitempty' flag.
This did not happen with my setup (Go 1.6.2).

brmasson:
I found this issue by Googling the error message: https://github.com/braintree/braintree_php/issues/137.
Does the merchant account used by Travis accept Maestro credit cards ?

brmasson:
@lionelbarrow The tests used to fail on Go 1.2 for reasons I did not understand but they seem to pass now. Is there anything I can do to improve my submission ?

Also would you consider removing support for older versions of Go in light of security issues with the standard crypto packages ? There are security fixes in Go releases as late as [1.7.5](https://github.com/golang/go/issues?q=milestone%3AGo1.7.5).

brmasson:
hello @leighmcculloch, I'll be happy to update that test

brmasson:
Note that the new field cannot be given to Braintree because it is part of the transaction [response object](https://developers.braintreepayments.com/reference/response/transaction/python#payment_instrument_type)

PR #80, opened 2016-08-19
Add DeviceData to transaction model with hyphen

leighmcculloch:
@lionelbarrow This PR takes the changes @cgcote opened in #76 and:

1. Changes the XML tag to be hyphenated rather than underscored.
2. Adds the field to the `TestAllTransactionFields` integration test using some fields specified [here](http://stackoverflow.com/a/34145834/159762). It's worth noting that adding it doesn't actually test anything, and there's no way to verify that the device data has been communicated correctly because it is not exposed when reading a `Transaction` from the API.

 I looked at the ruby SDK and it does not test the Device Data, so I think this is the best we can do with integration/external only tests.

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -1,8 +1,9 @@
 package braintree
 
 import (
-	"github.com/lionelbarrow/braintree-go/nullable"
 	"time"
+
+	"github.com/lionelbarrow/braintree-go/nullable"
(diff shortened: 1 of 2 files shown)


lionelbarrow:
Looks good!

PR #82, opened 2016-09-06
Added JSON and BSON structures

brianpowell:
Added JSON and BSON structures to make working with the JSON and BSON data match the other libraries and ease the differences between SDKs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Braintree Go
 
-[![GoDoc](https://godoc.org/github.com/lionelbarrow/braintree-go?status.svg)](http://godoc.org/github.com/lionelbarrow/braintree-go) [![Build Status](https://travis-ci.org/lionelbarrow/braintree-go.png?branch=master)](https://travis-ci.org/lionelbarrow/braintree-go)
+[![GoDoc](https://godoc.org/github.com/brianpowell/braintree-go?status.svg)](http://godoc.org/github.com/brianpowell/braintree-go) [![Build Status](https://travis-ci.org/brianpowell/braintree-go.png?branch=master)](https://travis-ci.org/brianpowell/braintree-go)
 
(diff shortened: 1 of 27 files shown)


leighmcculloch:
I don't think we'd want to add JSON and BSON tags to the structs as the package only uses XML to communicate with the Braintree API. If you're proposing this because you'd like to serialize one of the structs like `braintree.Transaction` to JSON in your application, you can do so by creating a struct in your app that has the same fields and casting `braintree.Transaction` to your struct. This has been possible since Go 1.8 even if the tags are different. This would allow you to copy the contents to a matching struct with JSON tags that you define and would be better as you'd get full control over how it serializes to JSON.

Example:
https://play.golang.org/p/9GUwR0N7LP

```go
package main

import (
	"encoding/json"
	"fmt"
)

type Transaction struct {
	Id string `xml:"id,omitempty"`
}

type JSONTransaction struct {
	Id string `json:"id,omitempty"`
}

func main() {
	txn := Transaction{Id: "s9df89"}
	
	jsonTxn := JSONTransaction(txn)
	
	output, _ := json.MarshalIndent(jsonTxn, "", "  ")
	fmt.Println(string(output))
}
```

wmiller848:
@leighmcculloch Agreed, I think your solution is the best for Merchants wanting to deserialize core structs out into other formats.

Although I don't entirely hate providing the support as languages like (ruby, python, and javascript essentially give this for free) others like (java, c#) do not have such features (that we provide).

Also in your PR

https://github.com/lionelbarrow/braintree-go/pull/82/files#diff-04c6e90faac2675aa89e2176d2eec7d8R45

You changed all the imports and README values to point at your fork, which we can't merge into here.

leighmcculloch:
Closing due to inactivity.

PR #83, opened 2016-09-13
Adds date range field support - settled-at, dispute-date, etc.

rkulla:
Adds search support for `settled_at`, `created_at`, `authorization_expired_at`, `authorized_at`, `failed_at, gateway_rejected_at`, `processor_declined_at`, `submitted_for_settlement_at`, `voided_at`, `disbursement_date`, `dispute_date` (and any other range fields added to the braintree api in the future should work automatically).

I used an empty interface for the RangeField attributes because they were just using floats, which didn't allow for strings. Also because xml doesn't allow tag fields to have the same names and to not break compatibility with people already using this api with float values.

### DIFF
--- a/search.go
+++ b/search.go
@@ -28,9 +28,9 @@ type TextField struct {
 
 type RangeField struct {
 	XMLName xml.Name
-	Is      float64 `xml:"is,omitempty"`
-	Min     float64 `xml:"min,omitempty"`
-	Max     float64 `xml:"max,omitempty"`
+	Is      interface{} `xml:"is,omitempty"`
+	Min     interface{} `xml:"min,omitempty"`
+	Max     interface{} `xml:"max,omitempty"`
 }
 
 type MultiField struct {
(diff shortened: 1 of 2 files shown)


search.go
```diff
@@ -28,9 +28,9 @@ type TextField struct {
 
 type RangeField struct {
 	XMLName xml.Name
-	Is      float64 `xml:"is,omitempty"`
-	Min     float64 `xml:"min,omitempty"`
-	Max     float64 `xml:"max,omitempty"`
+	Is      interface{} `xml:"is,omitempty"`
+	Min     interface{} `xml:"min,omitempty"`
+	Max     interface{} `xml:"max,omitempty"`
 }
```
leighmcculloch: To preserve concrete types, would it work to keep this as `float64`, and create a new struct `RangeDateField` with `Is`, `Min` and `Max` fields of type `time.Time`?

jszwedko: I'd be 👍  to using concrete types here.

rkulla:
Yeah that's doable, it will just require making a separate method called AddRangeDateField() for those fields. One minute I almost have it working I think...

leighmcculloch:
:tada:

Issue #38, opened 2015-05-06
Unable to distinguish between network errors and api errors?


antonholmquist:
I'm trying determine the failure reason of `bt.Customer().Find(customerID)`.

It returns an error (but not of any exported type) so I can't seem to distinguish between network errors, and the customer not actually existing.

The README states "If Braintree was able to process the request correctly, but was unable to fulfill it due to a semantic failure (such as the credit card being declined) then a `BraintreeError` type is returned.". The problem is that `BraintreeError` doesn't seem to be defined anywhere (just an unexported `braintreeError`), and it's not used for customer find anyways.

Please help me to shed some light on this. Thank you!

lionelbarrow:
Ah, that might be a design bug -- it seems that `BraintreeError` should be exported so that you can check if you've got one.

lionelbarrow:
Fixed in https://github.com/lionelbarrow/braintree-go/commit/41de65ad3d1e7f57fa62f5cfb9941ed1560ddd33

antonholmquist:
`BraintreeError` is now exposed, but `bt.Customer().Find(customerID)´ never returns it, so I still can't distinguish between network errors and the customer not actually existing. The error that is returned is of the unexported type `invalidResponseError`.

antonholmquist:
Shall I post this as a separate issue or will you re-open it?

By the way, do you know if there is any chance that Braintree will accept this as an official Go client in the future?

lionelbarrow:
I'll look into it more.

I'm not sure if this will become an official library any time soon, but I'll pass along the interest.

antonholmquist:
Any news on this?

lionelbarrow:
I took another look at the code this afternoon. As I see it, there are four failure cases to be concerned with:

* A networking failure happens and the request is unable to reach Braintree. It seems unlikely that this would return in a 404. A 502 or 503 is more likely.
* The request reaches Braintree but the URL is wrong. Perhaps we're hitting `GET /merchants/{merchant-id}/castomer` or something like that. In this case Braintree will respond with a 404.
* The request reaches Braintree, hits a valid URL, but the ID in a find request is wrong. For example if we're hitting `GET /merchants/my-merchant/customers/foo-customer` and `foo-customer` does not exist. In this case Braintree with also respond with 404.
[...]

PR #86, opened 2016-10-28
Introduced the httpError type with HTTP status exposed to client calls

michaljemala:
Fixes the #38, e.g.:

```
var someID string
...
customer, err := bt.Customer().Find(someID)
if err != nil {
    if e, ok := err.(braintree.APIError); ok && e.StatusCode() == http.StatusNotFound {
        fmt.Println("Customer not found")
        // TODO Do something, e.g. create a new customer
    } else {
        log.Fatal(err)
    }
}
...
```


braintree.go
```diff
@@ -116,6 +116,11 @@ func (g *Braintree) execute(method, path string, xmlObj interface{}) (*Response,
 		g.Logger.Printf("<\n%s", string(btr.Body))
 	}
 
+	err = btr.httpError()
+	if err != nil {
+		return nil, err
+	}
+
```
leighmcculloch: A side effect of putting `httpError` ahead of `apiError` is that if braintree includes error details for any of the status codes it won't get serialized into a `BraintreeError` any longer. Looking at the [Braintree Java SDK](https://github.com/braintree/braintree_java/blob/be892f293346a7d9a77bda750976cebdb1ce0c6b/src/main/java/com/braintreegateway/util/Http.java#L275-L300) there is at least one other case it seems where a message will be included. So it's probably safer to attempt to parse the response for a `BraintreeError` and only after that fails, if the status code is `> 299`, return your new `httpError` instead of the `fmt.Errorf`.

response.go
```diff
@@ -111,17 +111,33 @@ func (r *Response) apiError() error {
 		b.statusCode = r.StatusCode
 		return &b
 	}
-	if r.StatusCode > 299 {
-		return fmt.Errorf("%s (%d)", http.StatusText(r.StatusCode), r.StatusCode)
```
leighmcculloch: This is where I think we should keep the `> 299` and instead of returning the `fmt.Errorf`, return your `httpError(r.StatusCode)`.

response.go
```diff
@@ -111,17 +111,33 @@ func (r *Response) apiError() error {
...
 
+func (r *Response) httpError() error {
+	switch r.StatusCode {
+	case 200, 201, 422:
+		return nil
+	default:
+		return httpError(r.StatusCode)
+	}
+}
+
```
leighmcculloch: If you do the changes in the two comments above you should be able to drop this function.

leighmcculloch:
@michaljemala: I've pushed two commits to this PR to make the changes I talked about above, plus I've also changed the error message that `httpError` generates so it's identical to the error messages that were being returned before. This means anyone who is currently using the error string won't see any change.

PR #87, opened 2016-11-07
Add descriptor to transaction

knifeedge:


knifeedge:
Could someone look to see why the test is failing on travis? I test locally and descriptor shows up and the test passes, but it I seem to get a blank descriptor when running the travis test.

PR #89, opened 2016-12-02
Change ProcessorResponseCode type to nullable.NullInt64 (Option 1)

Invictus321:
The Processor response code is sometimes returned by the service as an empty tag, which causes the error: "strconv.ParseInt: parsing \"\": invalid syntax" when parsing 

The nullable.NullInt64 type circumvents this


leighmcculloch:
@Invictus321 Could you share a test / example that causes this to happen?

leighmcculloch:
I managed to figure out how to cause this to happen. If you create a Transaction that fails due to being gateway rejected it will get an empty processor response code. Any attempt to `Find` that transaction will fail unmarshaling. I added a test in #116 and #117 demonstrating.

I know of three ways we can solve this:

1. PR #89 - @Invictus321's solution here, which changes the fields type from `int` to `*nullable.NullInt64`.
  Pros/Cons: Will break devs builds, because changing the fields type from int to a struct will require changes. This is the reason the CI tests fail on this change, because our tests assume it's an int.

2. PR #116 - Change the fields type from `int` to a type with int as it's underlying type, and custom marshal/unmarshaling that supports empty string.
  Pros/Cons: Shouldn't break most builds, since the field is still an int under the covers comparison and switch operations using ints will continue to work as they do now (tests would continue to pass).

3. PR #117 - Change the fields type from `int` to `string`, which is the type given to the field in the other Braintree SDKs.
  Pros/Cons: Will break builds like 1, but is simpler than 1 and would bring this SDK inline with the official SDKs.

I'm leaning towards #116. Thoughts @lionelbarrow @jszwedko @wmiller848?

jszwedko:
#116 does seem like the least breaking change and what I would opt for for now. Just to clarify, would we ever see a processor response code that is literally 0 (as opposed to empty)?

#117 could make sense depending on if arithmetic of response codes (or numeric range conditions) are useful. If they are useful, #89 would be preferable long term. I'd recommend queuing one of these up for a breaking release.

leighmcculloch:
@jszwedko: Okay, I'll run with #116 for now. I'll close #89 and #117, I'm not sure we should use either long term. Long term in the next major version I think we should look at alternatives for the `nullable.NullInt64s` that offer the same behavior but are easier to use. I've found them difficult to use in tests.

PR #92, opened 2017-01-12
Implement payment method gateway

vially:
This PR builds on top of #44 in order to provide a generic payment method gateway which works with both credit cards and PayPal nonces (similar to the way the latest [Ruby SDK](https://developers.braintreepayments.com/reference/request/payment-method/create/ruby) works).

PS: @maxsz I hope you don't mind that I've used your work for this

### DIFF
--- a/credit_card.go
+++ b/credit_card.go
@@ -1,8 +1,12 @@
 package braintree
 
-import "time"
+import (
+	"encoding/xml"
+	"time"
+)
 
 type CreditCard struct {
+	XMLName                   xml.Name           `xml:"credit-card`
 	CustomerId                string             `xml:"customer-id,omitempty"`
 	Token                     string             `xml:"token,omitempty"`
 	PaymentMethodNonce        string             `xml:"payment-method-nonce,omitempty"`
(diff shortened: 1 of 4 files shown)


maxsz:
Thanks for working on this. 
@lionelbarrow what do you think?

vially:
@maxsz It seems some tests fail on go `1.3`, `1.4` and `1.5` but they run fine on the other go versions. Anyone has any idea why that might happen?

buu700:
Thanks for adding this @vially and @maxsz! I was just about to submit an issue about exactly this feature.

Any updates or ETA on when you might review this for merging @lionelbarrow?

leighmcculloch:
@vially: Did you want to resolve the conflicts on this now that `PayPalAccount`s exist in `master` and this can add the additional functionality to `PaymentMethodGateway`?

vially:
@leighmcculloch  I can do this tomorrow but feel free to work on it if you want.

vially:
@leighmcculloch I've rebased and fixed the conflicts but it seems some tests still fail. Could that be caused by the flaky tests mentioned in #100?

vially:
I've addressed your latest feedback. While adding tests for PayPal nonces I noticed the `find`, `update` and `delete` methods weren't working on `API version 3` so I've changed all methods (including `create`) to `API version 4`. Was there any particular reason why `create` was still using version 3 of the API?

PR #93, opened 2017-01-17
Unmarshal Addons and Discounts for a Plan

TamalSaha:
Ref:
https://github.com/braintree/braintree_java/blob/master/src/main/java/com/braintreegateway/Plan.java#L18

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -6,13 +6,15 @@ import (
 )
 
 type Plan struct {
-	XMLName               string              `xml:"plan"`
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Thanks @TamalSaha! I was going to drop a comment about adding unmarshalling tests, but I've gone ahead and done it in #105 which extends this PR because I had to make changes to the Braintree Sandbox account that's used for Travis to be able to test it.

This PR will be marked as merged automatically when #105 is merged.

PR #96, opened 2017-03-03
Add support for paypal payment method 

leighmcculloch:
This PR takes the changes made by @maxsz in #44, fixes one issue with api versions and changes the tests so that they use a fake nonce that's available in sandbox so that the test creates a payment method then attempts to find it, which means the test can run against any sandbox account without a paypal account needing to be manually created.


leighmcculloch:
Note: I've added the bare minimum needed for a payment method gateway to be able to use it to create a payment method using nonces in the PayPal Account tests. I wanted to keep the additional code in this PR as low as possible so I didn't flush it out entirely.

@vially in #92 has further flushed out the `PaymentMethod` and `PaymentMethodGateway` and I can pull that functionality into master after this is merged.

sdcoffey:
👍

PR #97, opened 2017-03-04
Add functions for handling default payment methods.

leighmcculloch:
What
===
Add functions and fields for handling multiple and default payment methods:

* Add `PayPalAccountOptions` with `MakeDefault` as a field for setting a `PayPalAccount` to be a `Customer`s default payment method.
* Add `IsDefault()` to the `PaymentMethod` interface.
* Add `PaymentMethods()` to `Customer`.
* Add `DefaultPaymentMethod()` to `Customer`.

Why
===
Now that a `Customer` has multiple payment method types (`CreditCard` and `PayPalAccount`) we need to be able to discover the default payment method, because `Customer` only exposes the default credit card.

### DIFF
--- a/credit_card.go
+++ b/credit_card.go
@@ -58,6 +58,10 @@ func (card *CreditCard) GetToken() string {
 	return card.Token
 }
 
+func (card *CreditCard) IsDefault() bool {
+	return card.Default
+}
+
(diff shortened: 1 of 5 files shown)


jszwedko:
👍

lionelbarrow:
👍

PR #100, opened 2017-03-04
Fix the flaky settlement summary test

leighmcculloch:
What
===
Sort the results that get compared at the end of the settlement summary
test, before they are compared.

Why
===
Sometimes the results come back in different orders. The order doesn't
matter, we just want to check that the items are the same and sorting
them will ensure the tests passes even if they come back in a different
order.


jszwedko:
👍

lionelbarrow:
Nice catch! 👍

PR #105, opened 2017-03-09
Unmarshal Addons and Discounts for a Plan (extended)

leighmcculloch:
This PR extends #93 adding tests to check that they are unmarshalled correctly within a `Plan`.

I associated the existing `test_add_on` and `test_discount` with `test_plan` so that when retrieving the plan we can test the presence and unmarshalling of the add on and discount.


leighmcculloch:
Changed `AddOns` and `Discounts`  back to being non-pointers, because I'd only made them pointers in-case we ever did updating of Plans, but according to the Braintree documentation Plans cannot be updated via the API and even if they supported it, and we decided to support it in this library, we'd probably do it with a separate PlanRequest object rather than with the Plan model.

PR #94, opened 2017-01-17
Add support for Addons and Discounts for Subscriptions.

TamalSaha:

PR #107, opened 2017-03-17
Add SubscriptionRequest for Create and Update

leighmcculloch:
What
===
Adds `SubscriptionRequest` for `Create` and `Update` of subscriptions in place of the existing `Subscription` model.

Why
===
It will allow us to separate out fields specifically for creating and updating subscriptions from those that we get back from the API.

Notes
===
The change was pulled from #94 as part of breaking it up. The change will be merged to `master` but the PR is opened against `subscription-tests` so that it's diff only shows the changes after `subscription-tests` is merged.


lionelbarrow:
👍 This makes sense.

PR #108, opened 2017-03-18
Add descriptor to Transactions and Subscriptions

leighmcculloch:
What
===
Add descriptor to `Transaction` and `Subscriptions`.

Why
===
So that a merchant can set and see the dynamic descriptor if their
account supports it.

Notes
===
The expectations for descriptors on `Transaction`s are not included in
the `TestAllTransactionFields` test because they will only pass if the
sandbox account is configured for dynamic descriptors but are instead
separated into their own test with an explanation on that tests
dependency.

This change is a rewrite of the part of #94 that added descriptors and
a part of getting that PR's logic committed piece meal.


leighmcculloch:
Ignoring the failure because it's an existing flaky subscription test that fails half of the time. Will visit it separately.

PR #109, opened 2017-03-20
Add modifications (addons and discounts) to subscriptions

leighmcculloch:
What
===
Add modifications (addons and discounts) to subscriptions.

Why
===
So that a merchant can add or change addons and discounts attached to subscriptions.

Notes
===
This PR is a continuation of the work discussed in #94 to break that PR up.

### DIFF
--- a/modification_request.go
+++ b/modification_request.go
@@ -1,0 +1,66 @@
+package braintree
+
+import (
+	"encoding/xml"
+)
+
+type ModificationRequest struct {
+	Amount                *Decimal `xml:"amount,omitempty"`
+	NumberOfBillingCycles int      `xml:"number-of-billing-cycles,omitempty"`
+	Quantity              int      `xml:"quantity,omitempty"`
+	NeverExpires          bool     `xml:"never-expires,omitempty"`
+}
+
+type AddModificationRequest struct {
+	ModificationRequest
+	InheritedFromID string `xml:"inherited-from-id,omitempty"`
+}
+
+type UpdateModificationRequest struct {
+	ModificationRequest
+	ExistingID string `xml:"existing-id,omitempty"`
+}
+
+type ModificationsRequest struct {
+	Add               []AddModificationRequest
+	Update            []UpdateModificationRequest
+	RemoveExistingIDs []string
+}
+
+func (m ModificationsRequest) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
+	e.EncodeToken(start)
+	if len(m.Add) > 0 {
+		if err := m.marshalXMLModifications(e, "add", m.Add); err != nil {
+			return err
+		}
+	}
+	if len(m.Update) > 0 {
+		if err := m.marshalXMLModifications(e, "update", m.Update); err != nil {
+			return err
+		}
+	}
+	if len(m.RemoveExistingIDs) > 0 {
+		if err := m.marshalXMLModifications(e, "remove", m.RemoveExistingIDs); err != nil {
+			return err
+		}
+	}
+	e.EncodeToken(start.End())
+	return nil
+}
+
+func (m ModificationsRequest) marshalXMLModifications(e *xml.Encoder, name string, modifications interface{}) error {
+	attr := []xml.Attr{{Name: xml.Name{Local: "type"}, Value: "array"}}
+	start := xml.StartElement{Name: xml.Name{Local: name}, Attr: attr}
+
+	if err := e.EncodeToken(start); err != nil {
+		return err
+	}
+	if err := e.EncodeElement(modifications, xml.StartElement{Name: xml.Name{Local: "modification"}}); err != nil {
+		return err
+	}
+	if err := e.EncodeToken(start.End()); err != nil {
+		return err
+	}
+
+	return nil
+}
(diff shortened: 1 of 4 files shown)


subscription_integration_test.go
```diff
@@ -745,3 +745,104 @@ func TestSubscriptionAllFieldsWithTrialPeriodNeverExpires(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestSubscriptionModifications(t *testing.T) {
```
leighmcculloch: I could flesh out this test so that it exercised Update and Remove, I'll push some additional tests for this.

modification_request.go
```diff
@@ -0,0 +1,66 @@
...
+	ExistingID string `xml:"existing-id,omitempty"`
+}
+
+type ModificationsRequest struct {
+	Add               []AddModificationRequest
+	Update            []UpdateModificationRequest
+	RemoveExistingIDs []string
+}
+
+func (m ModificationsRequest) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
```
leighmcculloch: The following 36 lines of code are to serialize the `ModificationsRequest` so that each of the `add`, `update` and `remove` tags within it have an attribute `type="array"` and to hide this as much as possible from the developer using the struct. The attribute is required otherwise the Braintree API returns a 403, even if the data within the tag is correct. I considered three approaches:

1. Add additional structs for `Add`, `Update`, `RemoveExistingIDs` that contains the respective arrays and an XMLAttr that serialized to `type` and that the developer had to manually set to the string `"array"`. This seemed undesirable since the developer had to set the field manually.
2. Add additional structs for `Add`, `Update`, `RemoveExistingIDs` that contains the respective arrays and a field that was a type we defined that automatically serialized itself as `type="array"` on it's enclosing struct. This seemed better since the developer did not have to set the field to anything and could not change the value that it was serialized to, but it was still undesirable because it exposes details about the serialization of the object to the developer unnecessarily.
3. Do what is done here and manually serialize the struct adding the attribute to each of the `add`, `update`, and `remove` tags. I think this is the solution we should go with, however it adds considerably more code to the repo. 36 lines is a lot for something that is largely builtin. It can be shortened a little with reflection but the code became less clear.

Any tips on how I can simplify this are much appreciated!

_Note: There is another approach I attempted where we do (2) but without the additional layer of structs. This isn't possible in Go right now because while tags can be flattened in the struct and then serialized out to a nested hierarchy, attributes cannot._

subscription_integration_test.go
leighmcculloch: This is done now. ✔️

modification_request.go
wmiller848: So the basic issue here is `encoding/xml` does not serialize `Add               []AddModificationRequest` with the `type="array"` property? With that being the case I agree with this approach of handling it under the hood.

subscription_integration_test.go
```diff
@@ -745,3 +745,136 @@ func TestSubscriptionAllFieldsWithTrialPeriodNeverExpires(t *testing.T) {
...
+						Amount:       NewDecimal(100, 2),
+						Quantity:     1,
+						NeverExpires: true,
+					},
+				},
+			},
+		},
+	})
+
+	t.Log("sub2", sub2)
```
wmiller848: Is there a reason we log these in the test?

leighmcculloch: Not really, I based it off the pattern used in existing tests in this file and in others so they were uniform. We should probably cleanup the logging so it's more meaningful. I'll leave that for another time.

modification_request.go
leighmcculloch: Yup that's the issue.

TamalSaha: @leighmcculloch , this definitely works. I wonder whether it will be simpler to use another struct like I did in my pr. In that struct you can set the Type="array". That struct has to be a public struct but you can put that in an internal pkg to make sure outside pkgs can't use it. I have not tried in code.
https://github.com/lionelbarrow/braintree-go/pull/94/files#diff-1dd96d11be5f0821e43a50897a292b4dR42

leighmcculloch: @TamalSaha: Ah, that would be simpler since the xml representation would still be described using tags on structs rather than a custom marshaller. I'll give that a go and see how it looks. Thanks for the feedback!

leighmcculloch:
@wmiller848 @TamalSaha: Take a look at the diff now. I've replaced the manual marshaling to XML with the use of structs. It's technically three lines of code longer, but I think it's easier to reason about.

TamalSaha:
LGTM! Thanks.

TamalSaha:
@leighmcculloch, can this be merged? Then I can delete my fork :)

leighmcculloch:
@wmiller848 @lionelbarrow @jszwedko: looking for 👍

subscription_integration_test.go
leighmcculloch: Actually, these exist so that on CI we have a list of all the subscriptions and transactions that were created so we can retroactively go and look them up. I just found this useful so think we should leave these in.

wmiller848:
👍

PR #110, opened 2017-03-21
Adds date range field support - settled-at, dispute-date, etc. (cont.)

leighmcculloch:
This PR takes the changes proposed in #83 and makes some changes to that proposal that I think are needed or beneficial for it to be merged. If this PR is merged it will automatically close and merge #83. 

Main changes to call out:

* The field is named TimeRange to exactly match the Go type that is being searched with.
* Added an integration test.
* Corrected the time format that was being marshaled to be `YYYY-MM-DDTHH:mm:ssZ`.
* A few other less critical changes to the xml marshaling logic.

@rkulla: for putting the functionality together 👏 .

### DIFF
--- a/search.go
+++ b/search.go
@@ -42,9 +42,12 @@ type RangeDateField struct {
 }
 
 func (d RangeDateField) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
+	start = start.Copy()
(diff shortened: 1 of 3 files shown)


wmiller848:
🍪

PR #111, opened 2017-03-21
Parallelize tests

leighmcculloch:
What
===
Mark all the tests that should be able to run in parallel as tests that can be run in parallel.

Why
===
Build time is ~47 seconds, but most tests are blocking on network calls. Running the tests in parallel across 15 routines seems like the sweet spot where the tests will complete in ~5-6 seconds on a single core CPU. The shorter test time should help make making changes less time consuming.

There is one or two tests that are likely to have issues from time to time with the parallelism, and I can keep an eye on them and see about either tweaking or making them run serially after the parallel tests if they aren't fixable.


wmiller848:
🍪 if it works :). Did you ever see any issues with this running against the gateway? I know we have issues running multiple CL against it for testing, none of these test seem to be dependent on seeded data in a particular state.

leighmcculloch:
@wmiller848: it works because 💚 from travis above. There is a test that fails from time to time when running parallel, but that's no different to when Travis runs our tests concurrently across all the Go versions or when multiple PRs are built at the same time.

PR #113, opened 2017-03-23
Fix timezone in expectations on dates from Braintree

leighmcculloch:
What
===
Change the expectation in the trial subscription tests that calculate
the first billing date so that the expected billing day number is based
on `US/Mountain` time, and allow others to override that with an
environment variable.

Why
===
The trial subscription tests are failing on the billing day of month for
a few hours each day and that suggests the expectation is being
calculated on a different timezone than Braintree is calculating it when
defining the billing day of month. It looks like the Braintree Java SDK
has tests like this and they use `US/Mountain` time [1] to calculate the
billing day of month. According to Braintree development docs [2] they
calculate it off the timezone of the account which in the case of the
account must be this. In the few tests I've run it seems to pass every
time I've used this timezone, so I think that is the timezone of the
test account.

Since the timezone is dependent on the timezone of the sandbox account
it makes sense to allow it to be configurable.

[1]
https://github.com/braintree/braintree_java/blob/8bde15986e9c94ce6611cdb3c7f0776f795df5a4/src/test/java/com/braintreegateway/integrationtest/SubscriptionIT.java#L141
[2]
https://developers.braintreepayments.com/guides/recurring-billing/create/ruby#subscription-days


test_values.go
```diff
@@ -17,4 +19,16 @@ var testGateway = New(
 	os.Getenv("BRAINTREE_PRIV_KEY"),
 )
 
+var testTimeZone = func() *time.Location {
+	tzName := os.Getenv("BRAINTREE_TIMEZONE")
+	if tzName == "" {
+		tzName = "US/Mountain"
```
jszwedko: This feels like an odd default. Maybe it should be UTC?

leighmcculloch: Agreed, but the account we test with on travis is setup with this timezone, and according to the Braintree docs it can't be changed without emailing Braintree Support.

leighmcculloch: We could default it to UTC, and set the variable to `US/Mountain` in the `.travis.yml`?

jszwedko: Aha, I see! Then yes, I think defaulting to `UTC` and setting `BRAINTREE_TIMEZONE` in the `.travis.yml` makes sense.

leighmcculloch:
@jszwedko Done!

PR #114, opened 2017-03-24
Restructure environments for easy overriding

leighmcculloch:
What
===
Change `Environment` to be a `struct` containing `BaseURL` instead of it
being a type with underlying type string and a `BaseURL()` method.

Why
===
The `BaseURL()` function only returns hardcoded URLs and cannot
be overriden. This prevents a developer using the library from changing
the BaseURL. Interest in doing this was expressed in PR #56.

Using an exported struct with exported fields makes it easy for a
developer to construct their own environment. The use of a struct also
lets us expand it with other parameters if needed without breaking
existing uses, which is important as `Environment` is part of the
exported interface of the package.

@cjrd @lionelbarrow: Let me know if this doesn't meet the end you 
were trying to get to in #56. Otherwise this should close #56.

### DIFF
--- a/braintree.go
+++ b/braintree.go
@@ -8,27 +8,7 @@ import (
 	"net/http"
 )
 
-type Environment string
-
-const (
-	Development Environment = "development"
-	Sandbox     Environment = "sandbox"
-	Production  Environment = "production"
-
-	LibraryVersion = "0.9.0"
-)
-
-func (e Environment) BaseURL() string {
-	switch e {
-	case Development:
-		return "http://localhost:3000"
-	case Sandbox:
-		return "https://sandbox.braintreegateway.com"
-	case Production:
-		return "https://www.braintreegateway.com"
-	}
-	panic(`invalid environment "` + e + `"`)
-}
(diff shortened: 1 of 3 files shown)


jszwedko:
👍

My main concern is that this appears to be backwards incompatible. I think you could preserve most (or all?) of the compatibility by keeping `BaseURL` as a method and using a different name for the struct field.

leighmcculloch:
@jszwedko: Good point. I could make the struct field `GatewayBaseURL`?

leighmcculloch:
@jszwedko: After talking to some go devs in gophers slack I went with making the field unexported and naming it `baseURL`  and adding a `NewEnvironment` function. We don't _need_ it to be unexported but the common pattern is if you are going to expose it via a function, don't expose the field as well. Keeps the exported interface smaller and less confusing I guess.

We could also rollback and go with just the field because even though BaseURL() was exported prior to this change, I doubt anyone would be using it since it was a fixed value.

I'm happy to go either of these two ways.

jszwedko:
:+1: I like this

Issue #42, opened 2015-08-06
Custom fields


DavidHuie:
Support for custom fields would be very helpful.

cjrd:
I would also find this useful and am very willing to submit a PR if you could provide a higher level overview of how to implement this change =)

leighmcculloch:
@DavidHuie @cjrd: See #115, it adds custom fields support to customers and transactions.

PR #115, opened 2017-03-26
Add custom fields to transactions and customers

leighmcculloch:
What
===
Add custom fields to transactions and customers.

Why
===
Requested in #42. Allows users of the package creating customers and
transactions to specify custom fields which are supported by the API.

Ref
===

 * [Braintree Docs on Custom Fields](https://articles.braintreepayments.com/control-panel/custom-fields)
 * [braintree/braintree_ruby](https://github.com/braintree/braintree_ruby) client library:
   * [xml/generator.rb](https://github.com/braintree/braintree_ruby/blob/48b5f13cd6c8213787f0a83fdbccdb71bf075273/lib/braintree/xml/generator.rb)
   * [xml/parser.rb](https://github.com/braintree/braintree_ruby/blob/48b5f13cd6c8213787f0a83fdbccdb71bf075273/lib/braintree/xml/parser.rb)

### DIFF
--- a/customfields/custom_fields.go
+++ b/customfields/custom_fields.go
@@ -1,0 +1,60 @@
+package customfields
+
+import (
+	"encoding/xml"
+	"io"
+	"strings"
+)
+
+// CustomFields is a string to string map of custom field names to values.
+// Ref: https://articles.braintreepayments.com/control-panel/custom-fields
+type CustomFields map[string]string
+
+type xmlField struct {
+	XMLName xml.Name
+	Value   string `xml:",chardata"`
+}
+
+var nameMarshalReplacer = strings.NewReplacer("_", "-")
+var nameUnmarshalReplacer = strings.NewReplacer("-", "_")
+
+// MarshalXML encodes the map of custom fields names to values with the name as an xml tag and the value as it's contents. Tag names have underscores replaced with hyphens, as required by the Braintree API.
+func (c CustomFields) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
+	if len(c) == 0 {
+		return nil
+	}
+
+	err := e.EncodeToken(start)
+	if err != nil {
+		return err
+	}
+
+	for k, v := range c {
+		tag := nameMarshalReplacer.Replace(k)
+		e.Encode(xmlField{
+			XMLName: xml.Name{Local: tag},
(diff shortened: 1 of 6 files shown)


customfields/custom_fields.go
```diff
@@ -0,0 +1,60 @@
...
+
+type xmlField struct {
+	XMLName xml.Name
+	Value   string `xml:",chardata"`
+}
+
+var nameMarshalReplacer = strings.NewReplacer("_", "-")
+var nameUnmarshalReplacer = strings.NewReplacer("-", "_")
+
+// MarshalXML encodes the map of custom fields names to values with the name as an xml tag and the value as it's contents. Tag names have underscores replaced with hyphens, as required by the Braintree API.
```
jszwedko: I'd recommend wrapping this line.

customfields/custom_fields.go
```diff
@@ -0,0 +1,60 @@
...
+		e.Encode(xmlField{
+			XMLName: xml.Name{Local: tag},
+			Value:   v,
+		})
+	}
+
+	return e.EncodeToken(start.End())
+}
+
+// UnmarshalXML decodes the tags within into a map of custom fields names to values with the xml tag is the name and the contents is it's value. Tag names have hyphens replaced with underscores.
```
jszwedko: Same here

customfields/custom_fields.go
```diff
@@ -0,0 +1,60 @@
...
+// CustomFields is a string to string map of custom field names to values.
+// Ref: https://articles.braintreepayments.com/control-panel/custom-fields
+type CustomFields map[string]string
+
+type xmlField struct {
+	XMLName xml.Name
+	Value   string `xml:",chardata"`
+}
+
+var nameMarshalReplacer = strings.NewReplacer("_", "-")
```
jszwedko: Why are these replacements needed?

leighmcculloch: The other SDKs do this explicitly. It might be a side effect of the other SDKs converting en-mass all underscore variables to hyphenated variables before XML serialization. It seems the control panel doesn't allow hyphens in custom fields and I guess that prevents overlap.

jszwedko: :+1:

PR #117, opened 2017-03-27
Change ProcessorResponseCode type to string (Option 3)

leighmcculloch:
What
===
Added a type for the ProcessorResponseCode on Transaction that
unmarshals itself and considers an empty XML tag as zero.

Why
===
In some situations, e.g. a gateway reject, the processor response code
field will be an empty string. The processor response code field on
transaction is an int, but the default xml decoder will error when
unmarshaling an empty string into an int.

This bug was reported in #89, which suggested solving the problem by
wrapping the value in nullable.NullInt64 like we have for similar
fields. I did this as a spike out of an alternative to #89. See the
discussion in the comments on #89 for three ways we can solve this
problem.


leighmcculloch:
Closing because of the discussion in #89.

PR #84, opened 2016-09-22
Add v.zero support

janpantel:
This adds the capability to authenticate against Braintree using a PayPal v.zero access token.

The corresponding implementations in the JAVA library can be found here:
### Access token gateway constructor

https://github.com/braintree/braintree_java/blob/7a36af1760392222ab371d610effa0e690dc878d/src/main/java/com/braintreegateway/BraintreeGateway.java#L72
### Decomposition of the token

https://github.com/braintree/braintree_java/blob/d2f00dfe725db07ef32810e42a11652f612d3516/src/main/java/com/braintreegateway/CredentialsParser.java#L36
### Constructing the header

https://github.com/braintree/braintree_java/blob/7a36af1760392222ab371d610effa0e690dc878d/src/main/java/com/braintreegateway/util/Http.java#L248
https://github.com/braintree/braintree_java/blob/7a36af1760392222ab371d610effa0e690dc878d/src/main/java/com/braintreegateway/util/Http.java#L294

PR #120, opened 2017-03-31
Add support for Access Tokens and lay the groundwork for other Credentials

leighmcculloch:
What
===
Add support for an access token as credentials, and prevent use of features that are not supported with an access token (e.g. webhooks).

Why
===
A change is proposed in #84 which does the same. It's simple with minimal changes and realistically might be all we need. But that change results in the Braintree struct containing a set of fields (`PublicKey`, `PrivateKey`, `AccessToken`) that won't be set all the time and if we continue to extend support for other credentials (e.g. [client ID and client secret](https://github.com/braintree/braintree_java/blob/7a36af1760392222ab371d610effa0e690dc878d/src/main/java/com/braintreegateway/CredentialsParser.java#L13-L34)) the struct will become more confusing to reason about from the outside. Additionally we end up with the `Environment` and `MerchantId` fields being set in the case of an API key and derived for the access token.

For the reasons above I think we could keep the interface on the struct cleaner, in code and in go docs, if we move the different types of credentials behind an internal interface `credentials` which exposes an authorization header and the other details.

The `credentials` interface type and field on the exported `Braintree` struct are intentionally unexported so that we can continue to iterate on them if needed without breaking the exported API.

Notes
===
This change includes two changes to the exported API:

* Remove the exported `PublicKey` and `PrivateKey` from the exported `Braintree` struct since neither will be set when other forms of credentials are used. These fields do not provide any functionality to the developer and only echo back the keys passed in.

* Replace the exported `Environment` and `MerchantId` from the exported `Braintree` struct with functions that return same, since both are now derived from credentials rather than being predefined or mutable values. As both are now functions of the same names any use of these fields should take a small amount of effort to replace with calling the respective functions.


doc.go
```diff
@@ -1,13 +1,43 @@
...
 
+Initializing
+
+Initialize it with API Keys:
+
+	braintree.New(Sandbox, merchantId, publicKey, privateKey)
+
+Initialize it with an Access Token:
+
+	braintree.NewWithAccessToken(accessToken)
```
jszwedko: It'd probably be good to call out here that webhooks are not supported with this type of credential.

jszwedko:
I like the interface that you've created here. It is unfortunate that it has to break API compatibility, but I couldn't come up with a sensible way to do it without.

In lieu, I think it'd be prudent to add a `CHANGELOG.md` where we can record all changes (including backwards incompatible ones) as well as publishing at least incompatible changes as release notes.

doc.go
leighmcculloch: I added a comment to the NewWithAccessToken function that when that is used webhooks are unsupported.

PR #121, opened 2017-04-01
Add Settlement methods for SandBox

TamalSaha:
Transaction gateway currently only supports setting transaction status to `settle`. This adds extra methods to set status to also `settlement_confirm`, `settlement_decline` and `settlement_pending`.

cc: @sadlil

### DIFF
--- a/testing_gateway.go
+++ b/testing_gateway.go
@@ -1,0 +1,42 @@
+package braintree
+
+// TestGateway exports actions only available in the sandbox environment.
+type TestingGateway struct {
+	*Braintree
+}
+
+// Settle changes the transaction's status to settle.
+func (g *TestingGateway) Settle(transactionID string) (*Transaction, error) {
+	return g.setStatus(transactionID, "settle")
+}
+
+// SettlementConfirm changes the transaction's status to settlement_confirm.
+func (g *TestingGateway) SettlementConfirm(transactionID string) (*Transaction, error) {
+	return g.setStatus(transactionID, "settlement_confirm")
+}
+
+// SettlementDecline changes the transaction's status to settlement_decline.
+func (g *TestingGateway) SettlementDecline(transactionID string) (*Transaction, error) {
+	return g.setStatus(transactionID, "settlement_decline")
+}
+
+// SettlementPending changes the transaction's status to settlement_pending.
+func (g *TestingGateway) SettlementPending(transactionID string) (*Transaction, error) {
+	return g.setStatus(transactionID, "settlement_pending")
+}
+
+func (g *TestingGateway) setStatus(transactionID, status string) (*Transaction, error) {
+	if g.Environment != Production {
+		resp, err := g.execute("PUT", "transactions/"+transactionID+"/"+status, nil)
+		if err != nil {
+			return nil, err
+		}
+		switch resp.StatusCode {
+		case 200:
+			return resp.transaction()
+		}
+		return nil, &invalidResponseError{resp}
+	} else {
+		return nil, &testOperationPerformedInProductionError{}
+	}
+}
--- a/transaction_gateway.go
+++ b/transaction_gateway.go
@@ -44,20 +44,9 @@ func (g *TransactionGateway) SubmitForSettlement(id string, amount ...*Decimal) 
 
 // Settle settles a transaction.
 // This action is only available in the sandbox environment.
+// Deprecated: use the Settle function on the TestingGateway instead. e.g. g.Testing().Settle(id).
 func (g *TransactionGateway) Settle(id string) (*Transaction, error) {
-	if g.Environment != Production {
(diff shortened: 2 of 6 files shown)


leighmcculloch:
👏 This is great.

I'm thinking that we probably shouldn't have had the `Settle` function on the `TransactionGateway` and we should have a `TestingGateway` similar to [TestingGateway.java](https://github.com/braintree/braintree_java/blob/7a36af1760392222ab371d610effa0e690dc878d/src/main/java/com/braintreegateway/test/TestingGateway.java). Now that we'll have many more sandbox-only functions it would keep the godoc documentation clearer since the `TransactionGateway` won't be 60% testing functions once we have all of them. What do you think about adding the new functions to a `TestingGateway` and we'll mark the `Settle` function as deprecated and copy it over to the `TestingGateway`?

TamalSaha:
I think that sounds great! I am going to update my pr.

TamalSaha:
@leighmcculloch I have split out the `TestingGateway`. Do you know why the one particular test fails on old GO versions?

testing_gateway.go
```diff
@@ -0,0 +1,42 @@
+package braintree
+
+// This action is only available in the sandbox environment.
+type TestingGateway struct {
+	*TransactionGateway
```
leighmcculloch: So I'd keep this as *Braintree.

testing_gateway.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// Settle settles a transaction.
+func (g *TestingGateway) Settle(id string) (*Transaction, error) {
```
leighmcculloch: I reckon given that this is no longer in the context of the transaction gateway, we should name the `id` variable `transactionID` to make it clear to others that's the type of ID we're expecting.

testing_gateway.go
```diff
@@ -0,0 +1,42 @@
...
+func (g *TestingGateway) SettlementConfirm(id string) (*Transaction, error) {
+	return g.setStatus(id, "settlement_confirm")
+}
+
+// Declined a settlements.
+func (g *TestingGateway) SettlementDecline(id string) (*Transaction, error) {
+	return g.setStatus(id, "settlement_decline")
+}
+
+// Declined a settlements.
```
leighmcculloch: This comment and the two above it should follow the go pattern of function name as the first word.
e.g. `// SettlementPending changes the transaction's status to settlement_pending.`

Ref: https://blog.golang.org/godoc-documenting-go-code

transaction_gateway.go
```diff
@@ -42,24 +42,6 @@ func (g *TransactionGateway) SubmitForSettlement(id string, amount ...*Decimal)
 	return nil, &invalidResponseError{resp}
 }
 
-// Settle settles a transaction.
-// This action is only available in the sandbox environment.
-func (g *TransactionGateway) Settle(id string) (*Transaction, error) {
```
leighmcculloch: I think we should keep this function because others may be using it, but have it call out to `g.Testing().Settle(id)`, and add a note to the last line of the comments above this function that say `// Deprecated: use the Settle function on the TestingGateway instead. e.g. g.Testing().Settle(id)`. Then we'll remove this function in the next major version.

The deprecated comment format comes from: https://blog.golang.org/godoc-documenting-go-code

transaction_integration_test.go
```diff
@@ -517,183 +475,6 @@ func TestTrxPaymentMethodNonce(t *testing.T) {
 	}
 }
 
-func TestTransactionCreateSettleAndFullRefund(t *testing.T) {
```
leighmcculloch: I think we should leave this function and the one below it here, because they are primarily concerned with testing transaction refund functionality, and only use the test-only settle endpoint to get the transaction into a state where it can be refunded.

I agree it makes sense to move the other two tests that are in this file since they seem primarily concerned with testing that the test-only settle functionality works.

testing_gateway.go
```diff
@@ -0,0 +1,42 @@
+package braintree
+
+// This action is only available in the sandbox environment.
```
leighmcculloch: This comment should start with the types name for consistency with godoc.
e.g. `TestGateway exports actions only available in the sandbox environment.`

leighmcculloch:
>Do you know why the one particular test fails on old GO versions?

@TamalSaha: Ah yup, if the `TestSettlementBatch` function fails you can ignore it. It's flaky. I haven't had a chance to make it more reliable yet. Sorry for the 🔴 😞 .

leighmcculloch:
Rebase on master and you'll get greener builds because #122 is merged.

TamalSaha:
@leighmcculloch PTAL. All your comments should be addressed.

leighmcculloch:
Thanks @TamalSaha and @sadlil!!! This is great.

PR #122, opened 2017-04-02
Fixed settlement batch summary unmarshaling.

leighmcculloch:
What
===
Change the unmarshaling of the settlement batch summary records
`MerchantAccountId` field to unmarshal from the XML tag
`merchant-account_id` to `merchant-account-id`, and change the
`TestSettlementBatch` test to test the fields in the record.

The rewrite subsequently makes the `TestSettlementBatch` test more
reliable and no longer flaky.

Why
===
The XML tag was incorrect for the `MerchantAccountId` field so the field
was always an empty string. The test didn't test this field and was
generally flaky and would fail even when the data was correct.

### DIFF
--- a/settlement.go
+++ b/settlement.go
@@ -4,7 +4,7 @@ type Record struct {
 	XMLName           string   `xml:"record"`
 	CardType          string   `xml:"card-type"`
 	Count             int      `xml:"count"`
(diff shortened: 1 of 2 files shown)


leighmcculloch:
@TamalSaha: This fixes up the test that you were seeing failing randomly.

Issue #72, opened 2016-05-23
Sample Webhook notification


chambo-e:
Hello,
Is there a way to generate a sample webhook notification like in other drivers ?
I can work on a PR if needed.
Thanks

leighmcculloch:
@chambo-e: There isn't one in this package, but did you want to open a PR adding a webhook testing gateway? The java library has one we could use as a base: https://github.com/braintree/braintree_java/blob/master/src/main/java/com/braintreegateway/WebhookTestingGateway.java

RichardKnop:
@chambo-e See my PR: https://github.com/lionelbarrow/braintree-go/pull/142

PR #123, opened 2017-04-02
fixed issue with multiple signatures

atishpatel:
new pull request for #75 @leighmcculloch

### DIFF
--- a/hmac.go
+++ b/hmac.go
@@ -41,23 +41,23 @@ func (h hmacer) verifySignature(signature, payload string) (bool, error) {
 	return hmac.Equal([]byte(expectedSignature), []byte(signature)), nil
 }
 
-func (h hmacer) getMatchingSignature(signaturePairs string) (string, string) {
+func (h hmacer) getMatchingSignature(signaturePairs string) (sig string, ok bool) {
 	pairs := strings.Split(signaturePairs, "&")
 	for _, pair := range pairs {
 		split := strings.Split(pair, "|")
 		if len(split) == 2 && split[0] == h.publicKey {
-			return split[0], split[1]
+			return split[1], true
 		}
 	}
-	return "", ""
+	return "", false
 }
 
 func (h hmacer) parseSignature(signatureKeyPairs string) (string, error) {
 	if !strings.Contains(signatureKeyPairs, "|") {
 		return "", SignatureError{"Signature-key pair does not contain |"}
 	}
-	publicKey, signature := h.getMatchingSignature(signatureKeyPairs)
-	if publicKey != h.publicKey {
+	signature, ok := h.getMatchingSignature(signatureKeyPairs)
+	if !ok {
 		return "", SignatureError{"Signature-key pair contains the wrong public key!"}
 	}
 	return signature, nil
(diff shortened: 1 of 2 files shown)


hmac.go
```diff
@@ -41,19 +41,26 @@ func (h hmacer) verifySignature(signature, payload string) (bool, error) {
...
+	return "", ""
+}
+
+func (h hmacer) parseSignature(signatureKeyPairs string) (string, error) {
+	if !strings.Contains(signatureKeyPairs, "|") {
+		return "", SignatureError{"Signature-key pair does not contain |"}
 	}
-	publicKey := split[0]
+	publicKey, signature := h.getMatchingSignature(signatureKeyPairs)
 	if publicKey != h.publicKey {
```
leighmcculloch: This line is a little odd to keep. The new `getMatchingSignature` function already matches it to the public key, but then we're double checking the match here again rather than just confirming we got a signature.

hmac.go
```diff
@@ -41,19 +41,26 @@ func (h hmacer) verifySignature(signature, payload string) (bool, error) {
...
 
-func (h hmacer) parseSignature(signatureKeyPair string) (string, error) {
-	if !strings.Contains(signatureKeyPair, "|") {
-		return "", SignatureError{"Signature-key pair does not contain |"}
+func (h hmacer) getMatchingSignature(signaturePairs string) (string, string) {
+	pairs := strings.Split(signaturePairs, "&")
+	for _, pair := range pairs {
+		split := strings.Split(pair, "|")
+		if len(split) == 2 && split[0] == h.publicKey {
+			return split[0], split[1]
```
leighmcculloch: This function returns both the public key and the signature, but we don't really need the public key after matching, we just need the signature to do the comparison. How about we return the signature plus an error that might be set if we couldn't match?

hmac.go
```diff
@@ -41,19 +41,26 @@ func (h hmacer) verifySignature(signature, payload string) (bool, error) {
...
 	}
-	split := strings.Split(signatureKeyPair, "|")
-	if len(split) != 2 {
-		return "", SignatureError{"Signature-key pair contains more than one |"}
+	return "", ""
+}
+
+func (h hmacer) parseSignature(signatureKeyPairs string) (string, error) {
+	if !strings.Contains(signatureKeyPairs, "|") {
+		return "", SignatureError{"Signature-key pair does not contain |"}
```
leighmcculloch: This validation is probably out of date I think. Taking a look at the [Braintree Java SDKs implementation of this](https://github.com/braintree/braintree_java/blob/8bde15986e9c94ce6611cdb3c7f0776f795df5a4/src/main/java/com/braintreegateway/WebhookNotificationGateway.java#L22-L45) it doesn't look like it does this check. Instead it validates that only valid characters are included and then iterates through ignoring entries that don't have the pipe.

leighmcculloch: The other SDKs don't have this, but I don't see why keeping it would be harmful. Retracting my comment above.

leighmcculloch:
@atishpatel: This PR was starting to get stale so I've made the changes and added tests as well. Thanks for submitting this change! 🎉

leighmcculloch:
Merging even though the build is red because the same tests are failing on `master`.

leighmcculloch:
Fixes #72

PR #124, opened 2017-04-02
Escrow Support

atishpatel:
I just rebased #67, but it showed a bunch of commits so I'm submitting a new pull request.


TESTING.md
```diff
@@ -87,4 +87,4 @@ Number of cycles:       For the duration of the subscription.
 
 #### Testing disbursement details
 
-In Braintree's sandbox environment, transactions do not disburse immediately. The implementation of disbursement details is tested in `disbursement_integration_test.go` using an already-disbursed transaction on the sandbox account associated with the Travis CI build. This test does not pass in other environments, and serves as a proof-of-concept. To make all of the tests pass, you will either need to comment out this test, or replace the values inside of it with the values from a disbursed transaction on your account. 
+In Braintree's sandbox environment, transactions do not disburse immediately. The implementation of disbursement details is tested in `disbursement_integration_test.go` using an already-disbursed transaction on the sandbox account associated with the Travis CI build. This test does not pass in other environments, and serves as a proof-of-concept. To make all of the tests pass, you will either need to comment out this test, or replace the values inside of it with the values from a disbursed transaction on your account.
```
leighmcculloch: I don't see any visible changes here. Can we revert the changes to this file?

transaction_integration_test.go
```diff
@@ -725,3 +726,185 @@ func TestTransactionCreateWithCustomFields(t *testing.T) {
...
+		Options: &TransactionOptions{
+			SubmitForSettlement: true,
+			HoldInEscrow:        true,
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	id := txn.Id
+	// _, err = escrow(id)
```
leighmcculloch: Can we remove this comment?

transaction_integration_test.go
```diff
@@ -725,3 +726,185 @@ func TestTransactionCreateWithCustomFields(t *testing.T) {
 		t.Fatalf("Returned custom fields doesn't match input, got %q, want %q", x, customFields)
 	}
 }
+
+func TestHoldInEscrowOnCreate(t *testing.T) {
+	testSubMerchantAccountId := getSubMerchantAccount(t)
```
leighmcculloch: Can we mark all these new tests as parallel, by adding `t.Parallel()` as the first line of each function? We run all the tests across 15 goroutines since the majority are integration tests and spend most of their time blocked on network calls.

transaction_integration_test.go
```diff
@@ -725,3 +726,185 @@ func TestTransactionCreateWithCustomFields(t *testing.T) {
...
+	amount := NewDecimal(6200, 2)
+	txn, err := testGateway.Transaction().Create(&Transaction{
+		Type:   "sale",
+		Amount: amount,
+		CreditCard: &CreditCard{
+			Number:         testCreditCards["visa"].Number,
+			ExpirationDate: "05/14",
+		},
```
leighmcculloch: I noticed you pulled out some code into functions at the end to DRY the code. Another thing you can do is instead of defining a credit card for each transaction, use `PaymentMethodNonce: FakeNonceTransactable`.

transaction_integration_test.go
```diff
@@ -725,3 +726,185 @@ func TestTransactionCreateWithCustomFields(t *testing.T) {
...
+	if err != nil {
+		return err
+	}
+	if txn.Status != "submitted_for_settlement" && txn.Status != "settling" && txn.Status != "settled" {
+		t.Fatalf("expected Status to be submitted_for_settlement, settling, or settled, was %s", txn.Status)
+	}
+	return nil
+}
+
+var subMerchantAccountID string
```
leighmcculloch: I like the idea of caching this so we don't create more sub-merchants unnecessarily. If these were running serially it might make a bit difference, but I'm not sure it will for us because we run the tests in parallel. If you run the tests like `go test -parallel 15` do you see much difference with the caching used or not?

leighmcculloch:
These changes were included in #186 and are now merged into master.

leighmcculloch:
The changes in this PR via #186 are in `v0.14.0` which has been released.

PR #129, opened 2017-04-14
Separate structs for Transaction request and response

leighmcculloch:
What
===
Add `TransactionRequest` to define the parameters needed for creating a
`Transaction`, instead of using `Transaction` for both the request and
response.

Why
===
The primary reason is so that we can introduce request fields that have
entirely different fields on the request and the response. e.g.
`risk-data`.

Aside from the primary reason most fields on `Transaction` are not
settable in the request and having a separate request struct more
clearly defines what a merchant can set on the request, and is more
consistent with other Braintree SDKs like the Java SDK which has a
request and response object for each model.

Notes
===
This is a significant breaking change, but we've made breaking changes
like this before as we're still pre-v1 and the exported API would
degrade quickly if we try and we continue to add new fields without
separating the request and response.

### DIFF
--- a/examples/transaction/server.go
+++ b/examples/transaction/server.go
@@ -42,7 +42,7 @@ func createTransaction(w http.ResponseWriter, r *http.Request) {
 		os.Getenv("BRAINTREE_PRIV_KEY"),
 	)
 
-	tx := &braintree.Transaction{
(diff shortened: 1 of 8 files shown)


lionelbarrow:
Yea, I agree that this is the most direct way to model the problem.

Issue #127, opened 2017-04-12
RiskData


no1youknowz:
Hi,

Moving some processing from PHP to Go.  

I'll provide some context...

I'm using RiskData per here:  https://developers.braintreepayments.com/reference/response/transaction/php#risk_data

Which is something like:

	$result = Braintree_Transaction::sale([
		...
	    'riskData' => Array(
			'customerIp' => $this->ci->input->ip_address(),
			'customerBrowser' => substr($_SERVER['HTTP_USER_AGENT'], 0, 254) // 255 Characters Max
		)
	]);

I've seen that the **Transaction** struct contains DeviceData, so that's customerBrowser sorted.

Where would I be able to submit the IP?  

Thanks

leighmcculloch:
This library doesn't support sending risk data yet, but we should add it.

Gotchas I can see with adding it is that the request and response fields for risk data are different, and we use the one struct for both. We should probably separate the two by adding `TransactionRequest` like how we did recently for `Subscription`. Then `TransactionRequest` would have the risk data with customer browser and ip and, and `Transaction` would continue to have the risk data struct it has now with the id and reason.

@no1youknowz: Are you up for making the change and opening a PR?

leighmcculloch:
@no1youknowz: See #130.

no1youknowz:
Thanks, will take a look at this over the weekend/monday.

Out of interest, in transaction.go there is:

    // TODO: not all transaction fields are implemented yet, here are the missing fields (add on demand)

I'm adding in currency-iso-code.

Is there anywhere I can add this into my own code, rather than appending the struct and then it getting removed whenever I update the library?

Thanks!

leighmcculloch:
@no1youknowz: I don't think there's a way to do what you described, but that list of fields are just the fields no one has gotten around to including. I've just opened PR 131 that adds it it in.

leighmcculloch:
@no1youknowz: risk data is now in the transaction request in `master`, added in #130.

PR #130, opened 2017-04-14
Add risk data request to transaction

leighmcculloch:
What
===
Add transaction request fields for risk data for customer browser and
customer ID.

Why
===
The fields aren't settable, and requested in #127.

Notes
===
Should be merged to `master` after #129.


lionelbarrow:
:nice:

PR #131, opened 2017-04-14
Add CurrencyISOCode to Transaction

leighmcculloch:
What
===
Add CurrencyISOCode to Transaction on deserialization.

Why
===
The field was listed in the list of fields we have not yet included in
the Transaction struct. Adding it to the supported list was requested in
issue #127 and it's simple enough for us to include.


leighmcculloch:
@no1youknowz: See above, `CurrencyISOCode` is now included in the `Transaction` struct.

lionelbarrow:
Retroactive 👍

Issue #133, opened 2017-04-22
Specify default http.Client timeout


dannyvankooten:
Hey,

While right now we can use `NewWithHttpClient` to pass in our own HTTP client, would it make sense to choose a sensible default for the `Timeout` field instead of using `net/http`'s default client (which has no timeout)?

I went through our codebase after reading [this post](https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779), so just dropping notes where I can. 

Thanks for the renewed attention for this package by the way. Using it in production and it's been great! Any chance Braintree will "officially" support this one?

leighmcculloch:
@dannyvankooten: Thanks that's a good catch. I think it makes sense for us to do as you say and provide a default client that isn't `net/http`s. Taking a look at the Braintree official SDKs for [Java](https://github.com/braintree/braintree_java/blob/1ea23ca33e9290ffb79ce458b0978e94737981e1/src/main/java/com/braintreegateway/Configuration.java#L139), [Ruby](https://github.com/braintree/braintree_ruby/blob/4247c3a9ead174815879c1062c9169c589aec24c/lib/braintree/configuration.rb#L92), and [Node](https://github.com/braintree/braintree_node/blob/aec33195246d2a09a0b7a0efcf11ab9eaf01819b/lib/braintree/config.js#L7) they all set the timeout to 60 seconds by default.

leighmcculloch:
@dannyvankooten: See #134.

dannyvankooten:
Whoa, that was fast. Looks good to me!

On April 23, 2017 7:46:27 AM GMT+02:00, Leigh McCulloch <notifications@github.com> wrote:
>@dannyvankooten: See #134.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/lionelbarrow/braintree-go/issues/133#issuecomment-296421382

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

leighmcculloch:
Change has been made to set a 60s timeout by default which is consistent with the official SDKs. It'll be included in the next release, v0.12.0.

PR #134, opened 2017-04-23
Add default timeout of 60 seconds.

leighmcculloch:
What
===
Add default `http.Client` to `Braintree` that has a timeout of 60
seconds.

Why
===
The default `http.Client` has no timeout and waits indefinitely.

Having no timeout is not great, because any connectivity or response
issue without feedback will cause the application using this package
to block forever, or at least a very long time. We should set a sensible
default.

The official Braintree SDKs for Java, Node and Ruby all use 60
seconds as their timeout.

Requested in #133.

Notes
===
It's always possible for developers using this package to set their own
http.Client and configure it with a timeout appropriate to their situation.
This change does not change that.

I'm not sure about keeping the tests that I added to prove out the
functionality. In total they introduce 70 seconds worth of tests to
the test suite, which almost doubles the amount of time the suite
takes to run.


knifeedge:
This is a great idea since I encountered the waiting indefinitely when I first used the library and realized that no timeoust were set out of the box. I would update the documentation to make mention of this default timeout as well.

braintree.go
```diff
@@ -18,14 +19,24 @@ const (
...
 )
 
+const defaultTimeout = time.Second * 60
+
+var defaultClient = &http.Client{Timeout: defaultTimeout}
+
 // New creates a Braintree with API Keys.
 func New(env Environment, merchId, pubKey, privKey string) *Braintree {
-	return &Braintree{credentials: newAPIKey(env, merchId, pubKey, privKey)}
+	return &Braintree{
```
jferm: I have not written Go seriously in a while so this may be convention, but calling `NewWithHttpClient` with `defaultClient` here would reduce duplication ever so slightly

leighmcculloch: 0c43865

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: go
 go:
-  - 1.2
```
leighmcculloch: @lionelbarrow: You cool with removing go1.2?

leighmcculloch:
@lionelbarrow

PR #136, opened 2017-04-24
Added a client interface to use with the NewWithHttpClient

knifeedge:
This will still preserve the old functionality since you can still pass in an http.Client, but allows you to pass another type that wraps the client like: https://github.com/sethgrid/pester/blob/master/main.go

### DIFF
--- a/braintree.go
+++ b/braintree.go
@@ -5,17 +5,26 @@ import (
 	"encoding/xml"
 	"errors"
 	"fmt"
+	"io"
 	"log"
 	"net/http"
+	"net/url"
 )
-
-const LibraryVersion = "0.10.0"
 
 type apiVersion int
 
+type HTTPClient interface {
+	Do(req *http.Request) (res *http.Response, err error)
+	Get(url string) (res *http.Response, err error)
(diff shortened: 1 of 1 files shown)


braintree.go
```diff
@@ -5,17 +5,26 @@ import (
...
+	PostForm(url string, values url.Values) (res *http.Response, err error)
+	Head(url string) (res *http.Response, err error)
+}
+
 const (
-	apiVersion3 apiVersion = 3
-	apiVersion4            = 4
+	apiVersion3    apiVersion = 3
+	apiVersion4               = 4
+	LibraryVersion            = "0.10.0"
```
leighmcculloch: The change moving the `LibraryVersion` looks unnecessary and groups unrelated identifiers together. Could we remove that from the PR diff?

knifeedge:
@leighmcculloch Thank you for the good feedback, I had not seen the comment from Brad Fitzpatrick on this which makes sense for not making a client interface. I had also seen https://github.com/hashicorp/go-retryablehttp and I should have researched this more. I'll check for errors when I call the various Braintree methods, and retry then.

Issue #139, opened 2017-04-25
How To Get Signature From Request


RichardKnop:
Hello, 

How can I get signature from `*http.Request` object in my handler where I want to parse a webhook notification like this:

```go
payload, err := ioutil.ReadAll(r.Body)
if err != nil {
	http.Error(w, "Error reading request body", http.StatusBadRequest)
	return
}

bt.WebhookNotification().Parse(signature, string(payload))
```

I assume `payload` is just the content of `HTTP POST` data but where is `signature`? In one of the headers?

leighmcculloch:
In the request body there will be two form parameters, bt_payload and bt_signature. If you use webhookGateway.Parse(r.FormValue("bt_signature"), r.FormValue("bt_payload")) that function will use the hmacer to verify the signature and will return a *WebhookNotification.

We could probably add a helper ParseRequest method that takes a http.Request and unpacked the two. It would make using this more straightforward I think.

RichardKnop:
@leighmcculloch Oh thanks. This is helpful :)

PR #140, opened 2017-04-26
Add ParseRequest to WebhookNotificationGateway

leighmcculloch:
What
===
Add `ParseRequest` to `WebhookNotificationGateway` that takes a webhook
notification `http.Request`.

Why
===
To simplify unpacking webhook notifications. Since most developers will
be using the stdlib http.Request struct, we can unpack the fields rather
than requiring them to do so.

Refs
===
https://github.com/lionelbarrow/braintree-go/issues/139#issuecomment-297053267

@RichardKnop

### DIFF
--- a/webhook_notification.go
+++ b/webhook_notification.go
@@ -6,6 +6,8 @@ import (
 )
 
 const (
+	Check = "check"
+
 	DisbursementWebhook                      = "disbursement"
 	DisbursementExceptionWebhook             = "disbursement_exception"
 	SubscriptionCanceledWebhook              = "subscription_canceled"
--- a/webhook_notification.go
+++ b/webhook_notification.go
@@ -6,8 +6,7 @@ import (
 )
 
 const (
-	Check = "check"
-
+	CheckWebhook                             = "check"
 	DisbursementWebhook                      = "disbursement"
 	DisbursementExceptionWebhook             = "disbursement_exception"
 	SubscriptionCanceledWebhook              = "subscription_canceled"
--- a/webhook_notification_test.go
+++ b/webhook_notification_test.go
@@ -26,7 +26,7 @@ func TestWebhookParseRequest(t *testing.T) {
 
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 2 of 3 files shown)


webhook_notification.go
```diff
@@ -6,6 +6,8 @@ import (
 )
 
 const (
+	Check = "check"
+
```
sdcoffey: Any reason for the \n? It might look cleaner adding grouping this with the others.

webhook_notification_test.go
```diff
@@ -2,9 +2,35 @@ package braintree
...
+	"io/ioutil"
+	"net/http"
+	"strings"
 	"testing"
 )
 
+func TestWebhookParseRequest(t *testing.T) {
+	t.Parallel()
+
+	gateway := New(Sandbox, "mid", "sz9g7zhxz8838v7h", "0c809a2d2e8f4e4c817900ff441c9554")
```
sdcoffey: I don't review PRs in this repo very often, so pardon if this a dumb q, but the other tests in this file use a `testGateway` which is constructed with env vars. Any reason to hardcode credentials here?

webhook_notification.go
leighmcculloch: Not really, I'll remove the new line. Thanks!

leighmcculloch: I'll also add the `Webhook` suffix to match.

webhook_notification_test.go
leighmcculloch: We use the `testGateway` for integration tests to hit an account we have configured with plans, and the needed merchant accounts. This is a unit test and the public/private key were used to create the test setup values. If we change the account we use for integration testing or roll it's keys it'd be best if we didn't have to update this test since it's unrelated to the integration account anyway.

The values I've used for the public and private keys I generated in sandbox by adding a new key to an account, sent a webhook to `requestb.in`, then destroyed the keys, so they don't point to an actual account.

webhook_notification.go
leighmcculloch: b502f11

Issue #143, opened 2017-04-28
AVS/CVV/Gateway/Tax Amount/Tax Exempt Transaction Fields


ccahoon:
Hello! Thanks for the great library. It's exciting to see so much great work being done on it!

I've got a branch based on a fairly old revision that implements a couple more transaction fields (and some alternative implementations of great work you've done since then). I haven't updated to the latest revision, but in a few months when I do, I'd be able to submit a pull request, but I wanted to ask first—would you prefer a PR at that point, or just a note that I'd use them if you added them?

leighmcculloch:
@ccahoon: A PR would be fantastic! Feel free to drop a list of requested fields (separate issues for each preferably), but if you open a PR there's a chance that what gets merged will align with your own solution.

ccahoon:
Great! In the process of writing the test, I've discovered (I think) that the sandbox account doesn't have AVS and CVV rules enabled. Do you think it'd be problematic for other tests to turn them on?

leighmcculloch:
@ccahoon: I've add a merchant account to the sandbox account the tests run against that has AVS and CVS turned on. All the rules are turned on just for this merchant account id.

Merchant Account ID: `avs-and-cvs`

Note: `cvs` was supposed to be `cvv` but I made a typo, and can't change it. 😭

ccahoon:
Looks like we've done all of the bits of this issue! Thanks so much for working with me on it.

leighmcculloch:
Thanks for kicking this off and contributing with PRs. It led to some great improvements to the package all round!

PR #144, opened 2017-04-28
Add Gateway Rejection Reason to Transaction

ccahoon:
Add Gateway Rejection Reason to Transaction.

### DIFF
--- a/gateway_rejection_reason.go
+++ b/gateway_rejection_reason.go
@@ -4,11 +4,11 @@ type GatewayRejectionReason string
 
 const (
 	GatewayRejectionReasonApplicationIncomplete GatewayRejectionReason = "application_incomplete"
-	GatewayRejectionReasonAVS                                          = "avs"
-	GatewayRejectionReasonAVSAndCVV                                    = "avs_and_cvv"
-	GatewayRejectionReasonCVV                                          = "cvv"
-	GatewayRejectionReasonDuplicate                                    = "duplicate"
-	GatewayRejectionReasonFraud                                        = "fraud"
-	GatewayRejectionReasonThreeDSecure                                 = "three_d_secure"
-	GatewayRejectionReasonUnrecognized                                 = "unrecognized"
+	GatewayRejectionReasonAVS                   GatewayRejectionReason = "avs"
+	GatewayRejectionReasonAVSAndCVV             GatewayRejectionReason = "avs_and_cvv"
+	GatewayRejectionReasonCVV                   GatewayRejectionReason = "cvv"
+	GatewayRejectionReasonDuplicate             GatewayRejectionReason = "duplicate"
+	GatewayRejectionReasonFraud                 GatewayRejectionReason = "fraud"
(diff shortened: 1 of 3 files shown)


ccahoon:
I'm moving these to gateway_rejection_reason.go. What do you think about the prefix?

gateway_rejection_reason.go
```diff
@@ -0,0 +1,14 @@
...
+
+const (
+	GatewayRejectionReasonApplicationIncomplete GatewayRejectionReason = "application_incomplete"
+	GatewayRejectionReasonAVS                                          = "avs"
+	GatewayRejectionReasonAVSAndCVV                                    = "avs_and_cvv"
+	GatewayRejectionReasonCVV                                          = "cvv"
+	GatewayRejectionReasonDuplicate                                    = "duplicate"
+	GatewayRejectionReasonFraud                                        = "fraud"
+	GatewayRejectionReasonThreeDSecure                                 = "three_d_secure"
+	GatewayRejectionReasonUnrecognized                                 = "unrecognized"
```
leighmcculloch: The only identifier that is typed in this list is the first one, the others are untyped consts. e.g. https://play.golang.org/p/sK63YlkNXk

transaction.go
```diff
@@ -69,13 +70,10 @@ type TransactionRequest struct {
 // TODO: not all transaction fields are implemented yet, here are the missing fields (add on demand)
 //
 // <transaction>
-//   <custom-fields>
-//   </custom-fields>
```
leighmcculloch: Cheers, thanks!

transaction_integration_test.go
```diff
@@ -241,6 +241,10 @@ func TestTransactionCreateWhenGatewayRejectedFraud(t *testing.T) {
 		t.Fatalf("Got status %q, want %q", txn.Status, "gateway_rejected")
 	}
 
+	if txn.GatewayRejectionReason != GatewayRejectionReasonFraud {
+		t.Fatalf("Got gateway rejection reason '%s', wanted 'fraud'", txn.GatewayRejectionReason)
```
leighmcculloch: If you replace `'%s'` with `%q` you'll get a quoted string out the other side, which is consistent with the other expectations in this test file. e.g. https://play.golang.org/p/bCG-ipotJv

gateway_rejection_reason.go
ccahoon: Nice catch! 

FWIW, I looked around to see what the current state of things was to mimic it-- do you think it's worth a potentially breaking change to go around and update those for consistency? main.go, for example, has the bug you pointed out here (only one of the constants is typed). subscription.go has the same convention format as this, but has a string type. That kind of thing. disbursement.go does not have the same naming convention as this and is untyped.

leighmcculloch: I've been meaning to fix up those other enums. I think we should fix them all and make them all typed. It's a breaking change but rather minor break, and until we reach v1 we've been allowing breaking changes.

ccahoon:
#143

PR #142, opened 2017-04-26
Added WebhookTestingGateway for testing purposes.

RichardKnop:
I have added a `WebhookTestingGateway` for testing purposes.

This is useful for people who want to write unit/integration tests to verify their application processes braintree webhook notifications correctly.

Example usage would be:

```go
package integration_test

import (
  "testing"
  "net/http/httptest"

  "github.com/lionelbarrow/braintree-go"
)

func TestMyWebhook(t *testing.T) {
  bt := braintree.New(
		braintree.Sandbox,
		"merchaint_id",
		"public_key",
		"private_key",
	)

  r, err := bt.WebhookTesting().Request(
    "POST",
    "http://1.2.3.4/webhook/",
    braintree.SubscriptionChargedSuccessfullyWebhook,
    "123", // id
  )
  if err != nil {
    t.Fatal(err)
  }

  // You can now send the payload and signature to your webhook handler
  // and test your application's busines logic

  w := httptest.NewRecorder()
  router.ServeHTTP(w, r)

  // assertions
}
```

PR #145, opened 2017-04-30
Tidy the WebhookTestingGateway

leighmcculloch:
What
===
Changed each subject function that generated XML to return just the
template, and have the `subjectXML` function generate the subject XML
using the template.

Why
===
Realized after we merged #142 that the templating was occurring in every
function, when each function could just return it's template instead,
which appears simpler as the templating logic occurs higher up the call
stack.

cc @RichardKnop


RichardKnop:
👍

PR #146, opened 2017-05-01
Add TaxExempt and TaxAmount to Transaction and TransactionRequest.

ccahoon:
I based their position in the struct on their position in https://developers.braintreepayments.com/reference/request/transaction/sale/ruby


transaction_integration_test.go
```diff
@@ -435,6 +439,9 @@ func TestAllTransactionFields(t *testing.T) {
 	if tx2.ShippingAddress.PostalCode != tx.ShippingAddress.PostalCode {
 		t.Fatalf("expected ShippingAddress.PostalCode to be equal, but %s was not %s", tx2.ShippingAddress.PostalCode, tx.ShippingAddress.PostalCode)
 	}
+	if tx2.TaxAmount.Cmp(tx.TaxAmount) != 0 {
+		t.Fatalf("expected Amount to be equal, but %s was not %s", tx2.Amount, tx.Amount)
```
ccahoon: Oops. Working on it.

ccahoon:
Rebased and fixed test error message.

ccahoon:
#143

transaction.go
```diff
@@ -24,6 +24,8 @@ type Transaction struct {
 	Customer                    *Customer                 `xml:"customer,omitempty"`
 	BillingAddress              *Address                  `xml:"billing,omitempty"`
 	ShippingAddress             *Address                  `xml:"shipping,omitempty"`
+	TaxAmount                   *Decimal                  `xml:"tax-amount,omitempty"`
+	TaxExempt                   bool                      `xml:"tax-exempt,omitempty"`
```
leighmcculloch: The `TaxExempt` field should probably be a `*nullable.Bool`, otherwise it won't be sent if it's `false`, and it will be `false` rather than not set if it's not sent by the server.

transaction_integration_test.go
```diff
@@ -435,6 +439,9 @@ func TestAllTransactionFields(t *testing.T) {
 	if tx2.ShippingAddress.PostalCode != tx.ShippingAddress.PostalCode {
 		t.Fatalf("expected ShippingAddress.PostalCode to be equal, but %s was not %s", tx2.ShippingAddress.PostalCode, tx.ShippingAddress.PostalCode)
 	}
+	if tx2.TaxAmount.Cmp(tx.TaxAmount) != 0 {
+		t.Fatalf("expected TaxAmount to be equal, but %s was not %s", tx2.TaxAmount, tx.TaxAmount)
+	}
```
leighmcculloch: Could we add an expectation for the value of the `TaxExempt` field in this test?

transaction_integration_test.go
```diff
@@ -707,3 +714,27 @@ func TestTransactionCreateWithCustomFields(t *testing.T) {
...
+
+	txn, err = testGateway.Transaction().Find(txn.Id)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !txn.TaxExempt {
+		t.Fatalf("Transaction did not return tax exempt")
+	}
+}
```
leighmcculloch: Could we add a test for if neither the `TaxExempt` or `TaxAmount` fields are set?

ccahoon:
Great catches! Will do tomorrow. Thanks for the careful reviews so far and
for patience.

transaction.go
ccahoon: I might be doing something wrong, but it seems that at least as of this API version, not sending tax exempt implies that the transaction is not tax exempt, and tax exempt seems to always been sent across the wire after a transaction is created. I'm happy to push the changes I have, but if my understanding is correct, I am not sure they are an improvement for usability.

transaction_integration_test.go
ccahoon: It looks like TaxAmount always gets sent down? With the *Decimal type, I end up with the value: &braintree.Decimal{Unscaled:0, Scale:0} even if no amount is set in the request.

transaction.go
leighmcculloch: Ah yeah, I see that now too, good catch. Since it always seems to be received, it'll be easier to use as a simple `bool`.

transaction_integration_test.go
leighmcculloch: I just ran the `TestTransactionCreateSubmitForSettlementAndVoid` test with the logger turned on that logs the XML sent and received, and this is what came back for the tax amount field:

```xml
...
<tax-amount nil="true"/>
<tax-exempt type="boolean">false</tax-exempt>
...
```

The XML field is there, but it has an attribute `nil="true"`.

This is probably a bug that the packages `Decimal` type (and other types) don't translate the field with `nil="true"` into a nil. Looking at some of the official SDKs they seem to translate the attribute to mean the field is nil. e.g. https://github.com/braintree/braintree_ruby/blob/master/lib/braintree/xml/parser.rb#L58-L59

I'd add the expectation here it's nil, and separately we can take a look at the `Decimal` type and see if it's safe to change it's behavior on `nil="true"`, and if so, fix that ahead of merging this PR. My motivation for suggesting this is if we merge it as it is and fix this later others may develop logic depending on getting back a zero Decimal rather than a nil Decimal.

leighmcculloch:
@ccahoon: I resolved the conflicts with `master` and replied to your comments. Let me know what you think.

leighmcculloch:
@ccahoon: I've made some changes to this PR taking a different approach to #157. The main problem with #157 is it makes backwards incompatible changes to how fields are expected to work, and it does this in a subtle-not-so-obvious way. If it was obvious, I'd be inclined to go that route, but I think #157 needs to be part of a larger story of overhauling the use of nil, null, omitempty, etc in this package if we want to move away from Null objects to nil ones.

Instead I've gone a safer route and added a `NullDecimal` that mirrors the `NullInt64`, `NullBool` types. It behaves the same way. It will always be instantiated and it won't be valid in the case the field is nil.

Let me know what you think, and if you think it's 👍  I can get this in the next version (`v0.12.0`).

transaction.go
```diff
@@ -43,6 +43,8 @@ type Transaction struct {
 	Customer                     *Customer                 `xml:"customer,omitempty"`
 	BillingAddress               *Address                  `xml:"billing,omitempty"`
 	ShippingAddress              *Address                  `xml:"shipping,omitempty"`
+	TaxAmount                    *NullDecimal              `xml:"tax-amount,omitempty"`
```
leighmcculloch: This could probably simply be a `NullDecimal` rather than a `*NullDecimal`. It makes little difference since it will always be instantiated. 💭

transaction_integration_test.go
```diff
@@ -867,3 +880,56 @@ func TestTransactionCreateWithCustomFields(t *testing.T) {
...
+	txn, err = testGateway.Transaction().Find(txn.Id)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !txn.TaxExempt {
+		t.Fatalf("Transaction did not return tax exempt")
+	}
+	if txn.TaxAmount.Valid {
+		t.Fatalf("Transaction returned a valid TaxAmont but is tax exempt")
```
ccahoon: *TaxAmount

transaction_integration_test.go
```diff
@@ -867,3 +880,56 @@ func TestTransactionCreateWithCustomFields(t *testing.T) {
...
+
+	txn, err = testGateway.Transaction().Find(txn.Id)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if !txn.TaxExempt {
+		t.Fatalf("Transaction did not return tax exempt")
+	}
+	if txn.TaxAmount.Valid {
```
ccahoon: Elsewhere you used `.Valid != true` -- do you want to use `.Valid != false` here?

transaction_integration_test.go
```diff
@@ -867,3 +880,56 @@ func TestTransactionCreateWithCustomFields(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	txn, err = testGateway.Transaction().Find(txn.Id)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if g, w := txn.TaxExempt, false; g != w {
```
ccahoon: Generally in this PR there are three different styles for testing these values. This isn't a big deal to me, but I wonder if it is worth unifying.

transaction.go
ccahoon: I believe I understand and agree with the reasoning behind making this both a pointer and a NullDecimal, and I think it might be worth warning/explaining to future users/maintainers why it came to be that way.

In addition, maybe it's worth explaining that this is always populated when one of these structs is created from an API call of some kind, so users can potentially omit the nil check if they feel safe doing so, and go straight to checking for Valid.

leighmcculloch:
🤔  I think the reason pointers and the null objects came about for two reasons:

1. Pointers – required to satify `omitempty` behavior (a) when we don't want to pass a field, since an empty struct won't be considered empty by `omitempty`. This is now partly irrelevant to the `Transaction` struct since for requests it was replaced with `TransactionRequest`.

2. NullX – required to satisfy the empty `nil="true"` XML elements (b) that the Braintree API returns. We technically do have another way to solve this that I demonstrated in #157.

---

@ccahoon: After reading your comments, I agree, there is value in cleaning up the null usage in the SDK, even if it's a breaking change. Even if we stick with the way it's done now it would be beneficial to define exactly why this pattern is being used.

---

The different directions I see we could go in:

1. Use NullX structs and remove pointers (e.g. `NullInt` instead of `*NullInt`) from response structs (e.g. `Transaction`).  
    
    This might be beneficial since the pointers are rarely nil since the Braintree API usually returns all of the fields with `nil="true"`, so there's probably not going to be a change in behavior for most users. It also removes the extra nil check that has to be done that @ccahoon pointed out. It means the `TransactionRequest` and `Transaction` structs diverge a little since the request class would still need pointers to satisfy the `omitempty`.
    
    Because we're retaining the NullX structs that are compatible with the sql Null structs, this might be easier for those who are taking the values straight to a database.

2. Use pointers only and remove the NullX structs everywhere (e.g. `*int` instead of `*NullInt`).  
    
    We can support this with #157 to strip out `nil="true"` XML elements so that fields are actually `nil` if the Braintree API hasn't returned a value. This would keep the `TransactionRequest` and `Transaction` more consistent since both would use a pointer to indicate the field is not set. This might be harder to upgrade to for those that are passing the values directly into a database since they'd need to wrap them in the sql Null structs manually.
    
    My understanding is that the null objects were added to handle the `nil="true"` empty XML elements that we can resolved with #157 but it's possible they've become useful for mapping onto databases. @lionelbarrow: do you have any insight? @jszwedko: you wrote the original Null structs, got any thoughts on this?
 4ee818e5009c8399271efeafee47d05b08885241

3. Continue to use both.
  The only value I know of in doing this option is backwards compatibility.

jszwedko:
:+1: that is a thoughtful consideration of the approaches.

I've personally gone back and forth on the pointer vs. `Null*` approach. I think both are valid, but that they shouldn't be combined (i.e. with things like via `*NullInt`). I'm having trouble remembering why I opted to do that since you end up with the worst of both worlds.

The library would definitely benefit from being consistent in usage though.

ccahoon:
In a greenfield situation, I think I'd prefer the pointer option. But..

I think the NullX option is the easier of two for the library and library consumers to transition to. I think it accomplishes what I want, particularly given @leighmcculloch's observation that the split between TransactionRequest and Transaction makes it easier to deal with the omitempty behavior. (Even if not, could we use that struct tag when unmarshaling to use .Valid == false to mean 'empty' on NullX structs?)

leighmcculloch:
@ccahoon: I want to make sure we're on the same page. The omitempty feature only works with marshaling, not unmarshaling, and when it comes to structs it only works with pointers. So if we go with `NullInt` for the fields within `Transaction`, we'll still need to do either `*NullInt` or `*int` for the fields on `TransactionRequest`. Is that what you're suggesting we do? If we're going to use `*int` on `TransactionRequest` I'm leaning towards doing the same on `Transaction`. The fields are then also consistent with other fields with other types like `*Address`.

ccahoon:
@leighmcculloch Thanks for helping me get on the same page. I follow, now. As I mentioned in https://github.com/lionelbarrow/braintree-go/pull/176, I think I'm in favor of the `*int` approach.

transaction_integration_test.go
leighmcculloch: I don't think we are completely consistent, but I changed this test to match the other test below it in b214e30.

leighmcculloch:
@ccahoon: I've updated this PR to be consistent with #176. Can you 👀  and 👍  and then I'll merge it.

PR #147, opened 2017-05-01
Add CVVResponseCode and constants to Transaction.

ccahoon:
I wasn't sure whether we wanted constants in this one-- I'm happy to pull them out, since they're not in the Java implementation. Also, unsure stylistically about the multiline comment in the constant file, but here's a draft. Feedback welcome.

### DIFF
--- a/cvv_response_code.go
+++ b/cvv_response_code.go
@@ -3,14 +3,22 @@ package braintree
 type CVVResponseCode string
 
 const (
-	CVVMatches      CVVResponseCode = "M" // The CVV provided matches the information on file with the cardholder's bank.
-	CVVDoesNotMatch CVVResponseCode = "N" // The CVV provided does not match the information on file with the cardholder's bank.
+	// The CVV provided matches the information on file with the cardholder's bank.
+	CVVResponseCodeMatches CVVResponseCode = "M"
+
+	// The CVV provided does not match the information on file with the cardholder's bank.
+	CVVResponseCodeDoesNotMatch CVVResponseCode = "N"
 
 	// The card-issuing bank received the CVV but did not verify whether it was correct.
 	// This typically happens if the processor declines an authorization before the bank evaluates the CVV.
-	CVVNotVerified CVVResponseCode = "U"
+	CVVResponseCodeNotVerified CVVResponseCode = "U"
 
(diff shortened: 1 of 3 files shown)


ccahoon:
#143

transaction_integration_test.go
```diff
@@ -250,6 +250,42 @@ func TestTransactionCreateWhenGatewayRejectedFraud(t *testing.T) {
...
+			CVV:            "200", // Should cause CVV does not match response
+		},
+	})
+
+	if err.Error() != "Gateway Rejected: cvv" {
+		t.Fatal(err)
+	}
+
+	txnID := err.(*BraintreeError).Transaction.Id
+	txn, err := testGateway.Transaction().Find(txnID)
```
leighmcculloch: Same comment as on #148. Do we need the find?

cvv_response_code.go
```diff
@@ -0,0 +1,16 @@
...
+	CVVMatches      CVVResponseCode = "M" // The CVV provided matches the information on file with the cardholder's bank.
+	CVVDoesNotMatch CVVResponseCode = "N" // The CVV provided does not match the information on file with the cardholder's bank.
+
+	// The card-issuing bank received the CVV but did not verify whether it was correct.
+	// This typically happens if the processor declines an authorization before the bank evaluates the CVV.
+	CVVNotVerified CVVResponseCode = "U"
+
+	CVVNotProvided              CVVResponseCode = "I" // No CVV was provided.
+	CVVIssuerDoesNotParticipate CVVResponseCode = "S" // The CVV was provided but the card-issuing bank does not participate in card verification.
+	CVVNotApplicable            CVVResponseCode = "A" // The CVV was provided but this type of transaction does not support card verification.
```
leighmcculloch: Same comment as on #148 that we should probably namespace these with the types name as the prefix. But interested to hear what you think as well as thoughts from @lionelbarrow.

ccahoon: I think it's a good idea. On it.

ccahoon:
I think I've addressed all of your comments. Please take another look!

leighmcculloch:
This looks great! Thanks @ccahoon!

PR #148, opened 2017-05-01
Add AVS Response code fields to Transaction, add constants.

ccahoon:
I followed the existing pattern for adding the avs-and-cvs merchant account mentioned here:  https://github.com/lionelbarrow/braintree-go/issues/143#issuecomment-298123332

Something about it feels weird-- maybe it's because the merchant account ID itself is so readable, it feels out of place. In any case, I don't think there's a problem, but I thought I'd mention that it felt a little funny.

If I didn't add the newline in the Transaction struct, gofmt would reindent everything. No existing struct type has any grouping in this manner, so I pretty much expect to remove the newline during feedback, but I wanted to keep the diff small for other feedback first.


avs_response_code.go
```diff
@@ -0,0 +1,21 @@
+package braintree
+
+type AVSResponseCode string
+
+const (
+	AVSMatches      AVSResponseCode = "M" // The postal code or street address provided matches the information on file with the cardholder's bank.
```
leighmcculloch: Part of me thinks we should name these fields with the prefix like the `GatewayRejectionReason` consts, so that would change these to be `AVSResponseCodeMatches`. It'll make it long for some, but I think the consistency will be beneficial. Thoughts?

cc @lionelbarrow

transaction_integration_test.go
```diff
@@ -250,6 +250,150 @@ func TestTransactionCreateWhenGatewayRejectedFraud(t *testing.T) {
...
+	if err == nil {
+		t.Fatal("Did not receive error when creating invalid transaction")
+	}
+
+	if err.Error() != "Gateway Rejected: avs" {
+		t.Fatal(err)
+	}
+
+	txnID := err.(*BraintreeError).Transaction.Id
+	txn, err := testGateway.Transaction().Find(txnID)
```
ccahoon: Agreed — I based it on the code added in this commit. I'll update that code as well https://github.com/Figure53/braintree-go/commit/300002fbbb52629b1c09d313b9996bd01639a00b#diff-1ec59924dc3de25f061095c8a4eefdc1R233.

avs_response_code.go
ccahoon: Agreed, doing that.

ccahoon:
Please take another look!

leighmcculloch:
@ccahoon: Looks like this has some conflicts, maybe because I merged the CVV PR first?

ccahoon:
Ah, probably. I'll rebase.

ccahoon:
Rebased and tidied the commit history a bit while I was at it.

leighmcculloch:
Thanks @ccahoon!

PR #149, opened 2017-05-09
Added SubscriptionId field to Transaction.

RichardKnop:
This PR request adds `SubscriptionId` field to `Transaction` object.

This is part of XML object together with `PlanID` when transaction is part of a subscription, e.g.:

```xml
          <plan-id>GBP_interval_1_month</plan-id>
          <subscription-id>dyvvpb</subscription-id>
``` 

It is useful to have access to this field when processing webhooks so you can easily tell which subscription this new transaction belongs to.

### DIFF
--- a/subscription_integration_test.go
+++ b/subscription_integration_test.go
@@ -894,3 +894,69 @@ func TestSubscriptionModifications(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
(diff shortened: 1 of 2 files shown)


leighmcculloch:
Very nice, thanks, this is great!!!

RichardKnop:
:)

PR #153, opened 2017-05-15
Add errcheck static analysis for CI runs.

leighmcculloch:
What
===
Add errcheck static analysis for CI runs.

Why
===
Not handling errors will mean we'll write code where errors are silently
ignored. The errcheck tool checks that all functions that return errors
collect the error to do something with it or check that the error is
intentionally ignored.

Why only on CI
===
errcheck is not very fast and can take a couple seconds to run. Running
it on CI only should strike a balance where missed errors can be caugt
before changes are accepted but errchecks slowness won't impact
iterating quickly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,23 @@
 language: go
-go:
-  - 1.2
-  - 1.3
-  - 1.4
-  - 1.5
-  - 1.6
-  - 1.7
-  - 1.8
-  - tip
+matrix:
+  include:
+  - go: 1.2
+  - go: 1.3
+  - go: 1.4
+  - go: 1.5
+  - go: 1.6
+  - go: 1.7
+  - go: 1.8
+    env: ERRCHECK=true
+  - go: tip
 
 go_import_path: github.com/lionelbarrow/braintree-go
+
+install:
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -14,4 +14,5 @@ go_import_path: github.com/lionelbarrow/braintree-go
 script:
   - source .default.env
   - go vet ./...
+  - CGO_ENABLED=0 errcheck ./...
```
leighmcculloch: `CGO_ENABLED=0` is required to prevent errcheck from utilizing gcc. Apparently this is a bug in the latest version of Go, since `CGO_ENABLED` is supposed to default to `0` but it's defaulting to `1`.

braintree.go
```diff
@@ -103,7 +103,7 @@ func (g *Braintree) executeVersion(method, path string, xmlObj interface{}, v ap
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
```
leighmcculloch: Explicitly ignoring the return error of the close on the response body, because we don't care if it fails because we're reading, not writing. We will have already fully read the body at the point we call close.

PR #154, opened 2017-05-16
Transaction Status Constants

RichardKnop:
Added constants for transaction states so they can be referred from code using this library.

This is very useful when integrating with Braintree in application code and we need to check transaction status and execute some logic based on different states.

It is much better to refer to constants in this library than to compare with strings. Less room for error as states are defined in one place.

Values taken from Ruby lib: https://github.com/braintree/braintree_ruby/blob/master/lib/braintree/transaction.rb#L32

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -7,27 +7,29 @@ import (
 	"github.com/lionelbarrow/braintree-go/nullable"
 )
 
+type TransactionStatus string
+
 const (
-	TransactionStatusAuthorizationExpired   = "authorization_expired"
-	TransactionStatusAuthorizing            = "authorizing"
-	TransactionStatusAuthorized             = "authorized"
-	TransactionStatusGatewayRejected        = "gateway_rejected"
-	TransactionStatusFailed                 = "failed"
-	TransactionStatusProcessorDeclined      = "processor_declined"
-	TransactionStatusSettled                = "settled"
-	TransactionStatusSettlementConfirmed    = "settlement_confirmed"
-	TransactionStatusSettlementDeclined     = "settlement_declined"
-	TransactionStatusSettlementPending      = "settlement_pending"
-	TransactionStatusSettling               = "settling"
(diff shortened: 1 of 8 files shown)


transaction.go
```diff
@@ -7,6 +7,23 @@ import (
...
+	TransactionStatusFailed                 = "failed"
+	TransactionStatusProcessorDeclined      = "processor_declined"
+	TransactionStatusSettled                = "settled"
+	TransactionStatusSettlementConfirmed    = "settlement_confirmed"
+	TransactionStatusSettlementDeclined     = "settlement_declined"
+	TransactionStatusSettlementPending      = "settlement_pending"
+	TransactionStatusSettling               = "settling"
+	TransactionStatusSubmittedForSettlement = "submitted_for_settlement"
+	TransactionStatusVoided                 = "voided"
+	TransactionStatusUnrecognized           = "unrecognized"
```
leighmcculloch: Do you think it's worth typing these constants to say a `type TransactionStatus string`? Or would that cause compile errors for code consuming the transaction status field on `Transaction` since it's type would change from `string` => `TransactionStatus`?

RichardKnop: It would work, Stripe does it in similar way: https://github.com/stripe/stripe-go/blob/master/sub.go#L7

RichardKnop: Also see: https://github.com/stripe/stripe-go/blob/master/sub/client.go#L12

RichardKnop:
@leighmcculloch OK, I have defined a new type for transaction states and using that.

If this is the approach you want to use, probably subscription states could be handled the same way.

transaction.go
leighmcculloch: It looks like a nice pattern to use, I'm just curious how much breakage it'd cause for code that takes the value and stores it in a string field since go won't auto cast it to a string.

RichardKnop: The code that is using status field now would need to either cast to string with `string()` function or with `fmt.Sprintf`. I think projects importing this library probably have it pinned to a specific commit.

PR #157, opened 2017-05-20
Make XML deserialization respect nil attribute

leighmcculloch:
What
===
Remove any element in the XML from the Braintree API that has the `nil="true"` attribute before the XML is deserialized into structs.

Why
===
Braintree's XML format adds the `nil="true"` attribute to elements that are intended to be nil and not have a default value. In the official Braintree SDKs it looks like values on things like `Transaction` are explicitly set to `null`/`nil` when the attribute is present.

As it stands the Go SDK ignores the attribute and we end up with default values. At this stage this only impacts one element on `Subscription`, but there are planning to add more fields (e.g. `tax_amount` in #146) that would be impacted. 

Why not
===
While this causes the code to more broadly interpret the XML more accurately adding this change could have unexpected side effects. This change is good for `tax_amount` (`Decimal`) as it isn't entirely correct to get back `$0` tax amount if you didn't set a tax amount on the `Transaction` and makes the SDK behave more like the official SDKs.  This is bad for `nullable.NullInt64` fields as today those are deserialized into `{Valid: false, Int64: 0}` and the standard use right now would be to check if `Valid == false` and this change would cause those values to be `nil` instead.

Alternatives
===
Alternatively we could add a `Valid` value to the `Decimal` struct.

cc @ccahoon


leighmcculloch:
A change I could make to this PR to make it safer is to only strip out elements matching certain names, so it only impacted the behavior of new fields.

ccahoon:
In your "Why Not" section, you mention nullable.NullInt64— I think there is a middle ground, where we change struct elements that are pointers to nullables to just the nullable types (which, I think, is their intended usage?).

So, the disadvantage changes to "a number of struct fields change types from pointers to nullables to the nullables themselves". But I think that, combined with your change here, it wouldn't be a big change for users of these structs.

leighmcculloch:
@ccahoon: Interesting. That's a decent idea, so basically we'd be making this change but making it in such a way that the nullable types still get constructed. I think we could get to the same end goal your proposing without removing the pointers by making it explicitly part of the unpacking. If the field being unpacked into is a nullable type, still unpack into it with `Valid: false`. What do you think?

leighmcculloch:
I couldn't find a way to differentiate between fields at this point and do this in a way that wouldn't break backwards compatibility and create two systems of identifying non-present values. In some cases we'd have Null structs and in others nil pointers. I think given that this needs more thought I'm going to close this PR and probably start thinking if we need to change the approach to non-present values for some distant release. In any case it'd be jarring to existing users so I want to approach it slowly.

The main reason this was opened was to facilitate the TaxAmount not being present in responses. For that I've created a `NullDecimal` type that mirrors the `NullInt64` type that is already used, and that pattern is at least more consistent with the package in it's current state. Read more about it here https://github.com/lionelbarrow/braintree-go/pull/146#issuecomment-310880573.

leighmcculloch:
Reopening in relevance to the discussion in #146.

leighmcculloch:
FYI @knifeedge, this change is being included in #176 instead of rolling it separately.

Issue #156, opened 2017-05-19
support context.Context in all remote APIs


jmhodges:
Passing contexts around for timeouts, cancellation, and other behaviors is now common and well-documented in Go land. It'd be really nice to have support for that in braintree-go. Even Context versions of the methods like how `database/sql` was updated would work.

leighmcculloch:
@jmhodges: Would you like to open a PR adding context to one area of the SDK which could be iterated on, and then we can adapt that pattern to the rest of the library?

jmhodges:
Can do!

jmhodges:
Did the AddonGateway and AddressGateway in https://github.com/lionelbarrow/braintree-go/pull/158

PR #158, opened 2017-05-20
Add context support to Addon and AddressGateway

jmhodges:
Updates #156


leighmcculloch:
That was fast! I'm wondering two things:

* How can we continue to support older versions of go? This will make the library go1.7+ only. Maybe we could split the things using Context into separate files that only build on version of go1.7+? This might be hard to do with the pattern in this PR where context is added to every function in the call stack as it would mean duplicating the full library.
* Separate to that the surface area of the library increase dramatically with two functions that do everything, one with Context and one without. 

---

To address both, would this pattern be idiomatic?

Add a function `WithContext` to `Braintree` that returns a `*Braintree` altered that has that `Context`. Calls to the library would then look like:

```go
// setup
g := braintree.New(...)

// on each request
g.WithContext(ctx).Transaction().Create(...)
```
or
```go
// setup
g := braintree.New(...)
tg := g.Transaction()

// on each request
tg.WithContext(ctx).Create(...)
```

This would result in less changes, and make usage completely opt-in in a similar fashion that you can do `.WithContext` on a `http.Request`. It'd also be much easier to split those additional functions out into files that are only compiled with go1.7+ using built in go build tags.

The idea isn't as elegant for users always using Context on every call, and I'm not sure how idiomatic it is.

What do you think?

leighmcculloch:
I chatted to @vcabbage on [Gophers Slack](https://gophers.slack.com/archives/C029RQSEE/p1495252180981640) and they pointed me to the godocs where they say to only use the first parameter pattern.

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

https://golang.org/pkg/context/#pkg-overview

So I guess we're stuck with having to make a couple choices:
- Do we reduce support to go1.7+ or use build tags to split code? Could be a lot of effort.
- Do we add context versions of each function or to keep the function surface area of the library down, just add context to existing functions and break backwards compatibility?

jmhodges:
To the first question, the other alternative is to use the golang.org/x/net/context import that Does The Right Thing for all versions of Go. It builds on older Go's and uses the stdlib's context on newer ones

I'm inclined towards preserving backwards incompatibility.

leighmcculloch:
Yeah I think we definitely need to support go1.6 because that's what Google App Engine runs on, and if we're going to support it then may as well support all existing versions.

jmhodges:
I've pushed up a commit that uses the x/net/context import path, but the build breaks because the current travis configuration doesn't do a `go get` (or equivalent). Is there a reason why? I can just add it, if not.

Alternatively, I guess I could copy the Context interface, but that seems a lil silly.

jmhodges:
Okay, I just added the `go get`. Now I'm running into issues where `context/ctxhttp` needs `Request.Cancel` but that doesn't show up til Go 1.5.

Can I just toss out the support for earlier versions of Go than 1.6?

jmhodges:
@leighmcculloch You have an opinion on the versions? I'm assuming you're the maintainer now.

leighmcculloch:
@jmhodges: I'm going to take a few more days to think about this. I'd like to find an elegant way we could get the best experience for those using go1.7+ (which means not adding external dependencies) and an uninterrupted experience for everyone else. The way I see that happening is if we do the following:

* use the `context` package for builds using go1.7+
* don't support `context` at all on builds using go1.6 or earlier
* don't have two versions of the library to support ^

jmhodges:
Ping

leighmcculloch:
@jmhodges: Haven't forgotten about this PR, need a bit more time.

leighmcculloch:
@jmhodges: I think reducing this package's support to Go 1.6+ from the next version onwards make sense. (cc @lionelbarrow)

lionelbarrow:
I think that makes sense. Go upgrades have been pretty stable and 1.6 is well over a year old at this point.

leighmcculloch:
I think this pattern looks grea. Once AppEngine moves on from go1.6 we can drop that version and replace `golang.org/x/net/context` with `context`.

@jmhodges: Are you up for taking it all the way through the SDK?

jmhodges:
Can do. Would you like a different PR for that that builds on this one, a different PR that includes this one, or for me to add that work to this PR?

leighmcculloch:
I think this PR works.

leighmcculloch:
I just saw [these comments](https://groups.google.com/forum/#!topic/google-appengine-go/sOg4eaEpst0) about AppEngine getting go1.8 in the coming weeks and go1.9 is planned for shortly after, I think we could aim for a min of go1.7 in the version of this package that this PR ships in. So I think you can safely reduce test runs in this PR to go1.7+ and use the stdlib `context` package.

leighmcculloch:
Go1.8 is in App Engine beta now: https://twitter.com/francesc/status/880102346382073861

leighmcculloch:
Go1.8 is GA in App Engine now: https://cloudplatform.googleblog.com/2017/10/announcing-Go-1-8-on-App-Engine-Standard-Environment.html

Given that 1.8 is the default now and the main reason we were still supporting 1.6 was App Engine, I think we can drop support for 1.6 and 1.7. It should make this change simpler since we can use the stdlib's `context`.

I think we _could maybe_ just add `context.Context` as the first parameter to existing functions instead of adding a second set of functions. It'd be a breaking change, but the upgrade path is simple for users and keeps the packages API slimmer. This is easier to do if we're not supporting older versions.

Thoughts @lionelbarrow @jmhodges ?

leighmcculloch:
As of #193 we don't need to support Go1.6.

I think we can also drop adding new functions with the `Context` suffix and add the context as the first parameter to existing functions, but I don't feel strongly about this.

@jmhodges are you up for updating this PR accordingly? Alternatively I've got bandwidth to close this out if you don't.

jmhodges:
Hey, yeah, I would love for you to take over. Thank you!

leighmcculloch:
Superseded by #194.

PR #159, opened 2017-05-31
Add payment-method-nonce to Customer

lquach:
To create a payment method along with a customer.
The Ruby documentation: https://developers.braintreepayments.com/reference/request/customer/create/ruby#customer-with-a-payment-method


leighmcculloch:
@lquach: Thanks! Could you add a test to `customer_integration_test.go` that creates a customer using one of the fake nonces?

lquach:
Test added.

leighmcculloch:
Nice! Thank you!

lquach:
@leighmcculloch thank you!

leighmcculloch:
Aside to this we should probably looking at splitting `Customer` into `Customer` (the data returned) and `CustomerRequest` (the data sent when creating/updating) like we recently did to `Transaction`, since it now has `PaymentMethodNonce` which is only used when creating the customer and not in the response.

PR #163, opened 2017-06-10
Add support for VenmoAccount payment methods

leighmcculloch:
What
===
Add support for VenmoAccount payment methods.

Why
===
Because it is supported in the official Braintree SDKs.


leighmcculloch:
`TestTransactionVenmoAccountDetails` fails with:
>`payment_method_nonce` does not contain a valid payment instrument type
>Venmo accounts are not accepted by this merchant account.

PR #164, opened 2017-06-10
Add support for ApplePay payment methods

leighmcculloch:
What
===
Add support for ApplePay payment methods.

Why
===
Because it is supported in the official Braintree SDKs.

### DIFF
--- a/apple_pay_card.go
+++ b/apple_pay_card.go
@@ -1,0 +1,59 @@
+package braintree
+
+import (
+	"encoding/xml"
+	"time"
+)
+
+type ApplePayCard struct {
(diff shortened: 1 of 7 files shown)


lionelbarrow:
@leighmcculloch Looks good - I wonder if there's a way to reduce the duplication in customer.go between these payment methods, but nbd.

leighmcculloch:
@lionelbarrow I can probably simplify the duplication in customer, as well as the `AllSubscriptions` functions too. I'll take a look at doing that after merging this, #163, and #165, then deal with it for all payment methods at once.

PR #166, opened 2017-06-14
Fix the settlement summary test

leighmcculloch:
What
===
Removed the code that calculated which settlement batch date the
transaction was supposed to be in, and replaced it with getting the
settlement batch's date from the transactions `settlement_batch_id`
field.

Why
===
Calculating the settlement batch date for the transaction was
unnecessary and unreliable. We were using the wrong timezone, and while
we could change it to use the correct timezone we can instead get the
date of the settlement batch from the settlement batch ID on the
transaction which is more reliable.

Note
===
The official documentation states that the format of the
settlement_batch_id could change at anytime, which could mean the date
would cease to exist within the ID, but the documentation at todays date
explicitly exposes the format as including the date as a prefix so I
think we can deal with changing the test if it does change in the
future.

### DIFF
--- a/settlement_integration_test.go
+++ b/settlement_integration_test.go
@@ -2,7 +2,6 @@ package braintree
 
 import (
 	"testing"
-	"time"
 )
 
 func TestSettlementBatch(t *testing.T) {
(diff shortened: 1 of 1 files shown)


leighmcculloch:
Merging this straight away because it fixes a test that's been repeatedly breaking. Post-merge comments are welcome and I'll address them if there are any.

PR #176, opened 2017-07-14
Replace *NullBool/*NullInt64 with *bool/*int and bool/int

leighmcculloch:
What
===
Replace `*NullBool` / `*NullInt64` with `*bool` / `*int` and `bool` / `int` as appropriate.

| Struct                     | Field                    | Current Type | New Type |
| -------------------------- | ------------------------ | ------------ | -------- |
| `CustomerSearchResult`     | `CurrentPageNumber`      | `*NullInt64` | `int`    |
|                            | `PageSize `              | `*NullInt64` | `int`    |
|                            | `TotalItems `            | `*NullInt64` | `int`    |
| `TransactionSearchResult ` | `CurrentPageNumber`      | `*NullInt64` | `int`    |
|                            | `PageSize `              | `*NullInt64` | `int`    |
|                            | `TotalItems `            | `*NullInt64` | `int`    |
| `DisbursementDetails `     | `FundsHeld `             | `*NullBool`  | `bool`   |
|                            | `Success `               | `*NullBool`  | `bool`   |
| `Plan `                    | `BillingDayOfMonth `     | `*NullInt64` | `*int`   |
|                            | `BillingFrequency `      | `*NullInt64` | `*int`   |
|                            | `NumberOfBillingCycles ` | `*NullInt64` | `*int`   |
|                            | `TrialDuration `         | `*NullInt64` | `*int`   |
|                            | `TrialPeriod `           | `*NullBool`  | `bool`   |
| `Subscription `            | `NeverExpires `          | `*NullBool`  | `bool`   |
|                            | `NumberOfBillingCycles ` | `*NullInt64` | `*int`   |
|                            | `TrialPeriod `           | `*NullBool`  | `bool`   |
| `SubscriptionRequest `     | `BillingDayOfMonth `     | `*NullInt64` | `*int`   |
|                            | `NeverExpires `          | `*NullBool`  | `*bool`  |
|                            | `NumberOfBillingCycles ` | `*NullInt64` | `*int`   |
|                            | `TrialPeriod `           | `*NullBool`  | `*bool`  |

Why
===
`*NullBool` and `*NullInt64` are cumbersome to use. To use the `Subscription` `NumberOfBillingCycles` bool variation as an example, the user of the struct must make two comparisons to safely inspect the value.

```go
type Subscription struct {
  ...
  NumberOfBillingCycles *NullInt64
  ...
}

subscription, _ := subscriptionGateway.Create(...)

if nx := subscriptions.NumberOfBillingCycles; nx != nil && nx.Valid {
  fmt.Println("Subscription has a number of billing cycles of", nx.Int64)
} else {
  fmt.Println("Subscription does not have number of billing cycles")
}
```

This change reduces the code above to:

```go
type Subscription struct {
  ...
  NumberOfBillingCycles *int
  ...
}

subscription, _ := subscriptionGateway.Create(...)

if nx := subscriptions.NumberOfBillingCycles; nx != nil {
  fmt.Println("Subscription has a number of billing cycles of", *nx)
} else {
  fmt.Println("Subscription does not have number of billing cycles")
}
```

This is better because:
 * The user needs only perform one check to determine if a value has been supplied.
 * The user does not need to know about a custom type, that is otherwise representing a built-in type.
 * Pointer types can be passed directly into `sql` queries to set `NULL` in nullable database columns.

Also, in many of the cases we can remove pointers and null all together.

@ccahoon @knifeedge @jszwedko @lionelbarrow


ccahoon:
FWIW, I prefer this approach. I think it's safe and fair to expect users of the library to understand pointers to these types and to have to think about when they'll be initialized.

leighmcculloch:
This is the solution I'm leaning towards too, although I've realized that looking at some of the fields, not all the fields we currently have marked with nullables are actually nullable. So before I merge this I plan to go through and change any to plain `int`s that do not need nullables.

leighmcculloch:
@ccahoon: Agreed, I think it's also fairer if we only place pointers on fields that really need them. I've changed about half of the fields to not be pointers based on this. I think this is in a state it could be merged now but really should get some eyes from people actively using this library first.

leighmcculloch:
@knifeedge: You mentioned to me at GopherCon that you had been thinking about this. Does the change in this PR look safe and beneficial from your POV?

knifeedge:
@leighmcculloch Sorry for the delay, I finally got time to look this over. I think the original intent of the Nullables (correct me if I am wrong) was to prevent xml errors when we received empty xml values and un-marshaling returned an error (with int and bool). I found this https://github.com/golang/go/issues/19333 where the plan for empty xml values is to return their zero value instead of getting back an error. The behavior of an empty xml field has been in-flux since 1.7 and in 1.9 it seems that empty values will return zero values, and invalid input like parsing "abc" into an int will return an error. Even with this fix, the library would have to specify that a user upgrade to 1.9 to use it, or detect the current go version and mimic what 1.9 will do internally.

leighmcculloch:
@knifeedge: Ah that's very interesting that Go1.9 will have the xml parsing relaxed so that empty strings parse into the default values. Regardless, this PR already handles this situation for all versions of Go. When Braintree a `nil` field back to us, the content of the element is not just empty. The element also has a `nil="true"` attribute. The first commit pre-parses the XML and strips out any elements that have a `nil="true"` attribute so that the Go parses that maps the XML to the struct won't get an empty element for `nil` fields. I based this design on the other Braintree SDKs that do a similar thing, where they skip over XML elements that have the `nil="true"` attribute.

knifeedge:
@leighmcculloch I did not realize that Braintree sent back a nil="true" attribute for some of the fields, another solution that Nullables provided. The current pull request is a much better improvement, the only suggestion would be to document the fields with the nil="true" attribute so that users know to check for the nil case.

leighmcculloch:
>document the fields with the nil="true" attribute

Any field that is a pointer maps to an element that can be nil in the API
response. I'll include in the changelog the table above that shows which
fields are changing.

PR #178, opened 2017-07-15
[RFC] Replace `*NullBool` and `*NullInt64` with `bool` and `int` and utilize default values

leighmcculloch:
What
===
Replace `*NullBool` and `*NullInt64` with `bool` and `int` and utilize default values for fields when they are not received, and default values to indicate that they should not be sent.

Why
===
`*NullBool` and `*NullInt64` are cumbersome to use. To use the `Subscription` `NumberOfBillingCycles` bool variation as an example, the user of the struct must make two comparisons to safely inspect the value.

```go
type Subscription struct {
  ...
  NumberOfBillingCycles *NullInt64
  ...
}

subscription, _ := subscriptionGateway.Create(...)

if nx := subscriptions.NumberOfBillingCycles; nx != nil && nx.Valid {
  fmt.Println("Subscription has a number of billing cycles of", nx.Int64)
} else {
  fmt.Println("Subscription does not have number of billing cycles")
}
```

This change reduces the code above to:

```go
type Subscription struct {
  ...
  NumberOfBillingCycles int
  ...
}

subscription, _ := subscriptionGateway.Create(...)

if nx := subscriptions.NumberOfBillingCycles; nx != 0 {
  fmt.Println("Subscription has a number of billing cycles of", *nx)
} else {
  fmt.Println("Subscription does not have number of billing cycles")
}
```

This is better because:
 * The user needs only perform one check to determine if a value has been supplied.
 * The user does not need to know about a custom type, that is otherwise representing a built-in type.
 * Using default values is simple, and in many cases meaningful. If we do not receive many boolean or int fields, the meaning for the zero value or no value is the same.

However, there are downsides:
 * We may have edge cases that are difficult to see where there is a difference between the zero value and the no value. These differences may be significant, or insignificant.

Alternatives
===
See #177, #176.

Notes
===
This PR is built upon #177, but that's just happenstance, if it is picked it'd be squashed so that the proposal in #177 wasn't a part of this PR.

Merging
===
This change is dependent on #157 and is intended to be merged to `master` after #157.

@ccahoon @knifeedge @jszwedko @lionelbarrow


leighmcculloch:
Going to reopen this in smaller slices and with a better git history.

PR #182, opened 2017-08-04
Add transaction options for paypal request

bastiankoetsier:
This is basically the port from the official  [Java-Library](https://github.com/braintree/braintree_java/blob/7a36af1760392222ab371d610effa0e690dc878d/src/main/java/com/braintreegateway/TransactionOptionsPayPalRequest.java) to support sending PayPal-specific options while creating a transaction.

Let me know if there is missing something 😄  Thanks for your awesome work!

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -155,34 +155,54 @@ type TransactionOptions struct {
 }
 
 type TransactionOptionsPaypalRequest struct {
-	CustomField       string               `xml:"custom-field,omitempty"`
-	PayeeEmail        string               `xml:"payee-email,omitempty"`
-	Description       string               `xml:"description,omitempty"`
-	SupplementaryData supplementaryDataMap `xml:"supplementary-data,omitempty"`
+	CustomField       string
+	PayeeEmail        string
+	Description       string
+	SupplementaryData map[string]string
 }
 
-type supplementaryDataMap map[string]string
+func (r TransactionOptionsPaypalRequest) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
+	type transactionOptionsPaypalRequest TransactionOptionsPaypalRequest
 
-func (s supplementaryDataMap) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
-
-	tokens := []xml.Token{start}
-
-	for key, value := range s {
-		t := xml.StartElement{Name: xml.Name{Space: "", Local: key}}
-		tokens = append(tokens, t, xml.CharData(value), xml.EndElement{Name: t.Name})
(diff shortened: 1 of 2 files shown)


transaction.go
```diff
@@ -146,10 +146,18 @@ type Transactions struct {
...
+	AddBillingAddressToPaymentMethod bool                             `xml:"add-billing-address-to-payment-method,omitempty"`
+	StoreShippingAddressInVault      bool                             `xml:"store-shipping-address-in-vault,omitempty"`
+	TransactionOptionsPaypalRequest  *TransactionOptionsPaypalRequest `xml:"paypal,omitempty"`
+}
+
+type TransactionOptionsPaypalRequest struct {
+	CustomField       string            `xml:"custom-field,omitempty"`
+	PayeeEmail        string            `xml:"payee-email,omitempty"`
+	Description       string            `xml:"description,omitempty"`
+	SupplementaryData map[string]string `xml:"supplementary-data,omitempty"`
```
leighmcculloch: Could you add this field to the test?

bastiankoetsier:
It is a good thing you pointed out the missing fields in the test: Turns out you cannot use a "plain" map for XML marshalling :-/ ( I don't use XML very often 🤷‍♂️ )  So, I added a custom type that supports this. Unfortunately, as you can see here in the [official PHP lib](https://github.com/braintree/braintree_php/blob/0614d169e3ef8a9b16a90367f8a66bdd9f5a9df6/tests/integration/TransactionTest.php#L3313) the supplementary data is not returned in the response, so I cannot assert against it :-(

bastiankoetsier:
Hey @leighmcculloch , 
this does not require any more work from my side, does it? It seems that some tests are very flaky against the sandbox and they do not seem related to my changes 😐

beaspider:
+1 for including this in the v0.13.0 release

transaction.go
```diff
@@ -146,10 +147,46 @@ type Transactions struct {
...
+	PayeeEmail        string               `xml:"payee-email,omitempty"`
+	Description       string               `xml:"description,omitempty"`
+	SupplementaryData supplementaryDataMap `xml:"supplementary-data,omitempty"`
+}
+
+type supplementaryDataMap map[string]string
+
+func (s supplementaryDataMap) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
+
+	tokens := []xml.Token{start}
```
leighmcculloch: Can we drop the intermediary `tokens` slice, and encode each token at each step?

leighmcculloch:
@bastiankoetsier: I'll make these changes this arvo and see if I can get it in the v0.13.0 release today.

PR #183, opened 2017-08-17
Change TransactionStatus type to string type to eliminate unnecessary type assertions

ulisesflynn:
When the constants were added in [#154](https://github.com/lionelbarrow/braintree-go/pull/154) they could have been of type string versus TransactionStatus. The type doesn't provide any extra functionality, and now requires an unneeded type assertion.

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -6,29 +6,27 @@ import (
 	"github.com/lionelbarrow/braintree-go/customfields"
 )
 
-type TransactionStatus string
-
 const (
-	TransactionStatusAuthorizationExpired   TransactionStatus = "authorization_expired"
-	TransactionStatusAuthorizing            TransactionStatus = "authorizing"
-	TransactionStatusAuthorized             TransactionStatus = "authorized"
-	TransactionStatusGatewayRejected        TransactionStatus = "gateway_rejected"
-	TransactionStatusFailed                 TransactionStatus = "failed"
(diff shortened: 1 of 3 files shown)


transaction.go
```diff
@@ -6,29 +6,27 @@ import (
...
+	TransactionStatusFailed                 string = "failed"
+	TransactionStatusProcessorDeclined      string = "processor_declined"
+	TransactionStatusSettled                string = "settled"
+	TransactionStatusSettlementConfirmed    string = "settlement_confirmed"
+	TransactionStatusSettlementDeclined     string = "settlement_declined"
+	TransactionStatusSettlementPending      string = "settlement_pending"
+	TransactionStatusSettling               string = "settling"
+	TransactionStatusSubmittedForSettlement string = "submitted_for_settlement"
+	TransactionStatusVoided                 string = "voided"
+	TransactionStatusUnrecognized           string = "unrecognized"
```
leighmcculloch: If we're removing the type from these constants, we may as well make them untyped instead of `string`. Simply remove the `string` part of each line and it will be an untyped string.

ulisesflynn: The typed constants were making it so that we would have to type assert without gaining anything else. I've removed the unneeded string from the constants.

ulisesflynn:
Sorry for the delay for looking at this, my intent was not to have to do any casting, but I see that I could just use the TransactionStatus type from my local code.

leighmcculloch:
Great news, thanks @ulisesflynn !

PR #185, opened 2017-09-15
Add Clone method to TransactionGateway.

dsmcfarl:
Also added TransactionCloneRequest and supporting test.

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -98,27 +98,10 @@ type TransactionRequest struct {
 }
 
 type TransactionCloneRequest struct {
-	XMLName            string                    `xml:"transaction-clone"`
-	CustomerID         string                    `xml:"customer-id,omitempty"`
-	Type               string                    `xml:"type,omitempty"`
-	Amount             *Decimal                  `xml:"amount"`
-	OrderId            string                    `xml:"order-id,omitempty"`
-	PaymentMethodToken string                    `xml:"payment-method-token,omitempty"`
-	PaymentMethodNonce string                    `xml:"payment-method-nonce,omitempty"`
-	MerchantAccountId  string                    `xml:"merchant-account-id,omitempty"`
-	PlanId             string                    `xml:"plan-id,omitempty"`
-	CreditCard         *CreditCard               `xml:"credit-card,omitempty"`
-	Customer           *Customer                 `xml:"customer,omitempty"`
-	BillingAddress     *Address                  `xml:"billing,omitempty"`
-	ShippingAddress    *Address                  `xml:"shipping,omitempty"`
-	TaxAmount          *Decimal                  `xml:"tax-amount,omitempty"`
-	TaxExempt          bool                      `xml:"tax-exempt,omitempty"`
-	DeviceData         string                    `xml:"device-data,omitempty"`
-	Options            *TransactionOptions       `xml:"options,omitempty"`
-	ServiceFeeAmount   *Decimal                  `xml:"service-fee-amount,attr,omitempty"`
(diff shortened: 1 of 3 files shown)


transaction_integration_test.go
```diff
@@ -54,6 +54,25 @@ func TestTransactionCreateSubmitForSettlementAndVoid(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if x := tx4.Status; x != TransactionStatusSubmittedForSettlement {
+		t.Fatal(x)
+	}
+	if amount := tx4.Amount; amount.Cmp(ten) != 0 {
+		t.Fatalf("transaction settlement amount (%s) did not equal amount requested (%s)", amount, ten)
+	}
+
```
leighmcculloch: This test is focusing purely on submitting for settlement and then voiding. Can you move the expectations you've added here into a new test that is focused on creating a transaction and cloning it?

transaction.go
```diff
@@ -97,6 +97,30 @@ type TransactionRequest struct {
...
+	MerchantAccountId  string                    `xml:"merchant-account-id,omitempty"`
+	PlanId             string                    `xml:"plan-id,omitempty"`
+	CreditCard         *CreditCard               `xml:"credit-card,omitempty"`
+	Customer           *Customer                 `xml:"customer,omitempty"`
+	BillingAddress     *Address                  `xml:"billing,omitempty"`
+	ShippingAddress    *Address                  `xml:"shipping,omitempty"`
+	TaxAmount          *Decimal                  `xml:"tax-amount,omitempty"`
+	TaxExempt          bool                      `xml:"tax-exempt,omitempty"`
+	DeviceData         string                    `xml:"device-data,omitempty"`
+	Options            *TransactionOptions       `xml:"options,omitempty"`
```
leighmcculloch: Comparing the `TransactionsOptions` struct to the [TransactionOptionsCloneRequest](https://github.com/braintree/braintree_java/blob/acc7c9ada19e09f6181dd501657fc7f4f1b7afcb/src/main/java/com/braintreegateway/TransactionOptionsCloneRequest.java) in the Braintree Java SDK, I see that many of the fields in the `TransactionsOptions` struct are not intended to be in a clone options request. 

Can you create a new struct `TransactionCloneOptions` similar to `TransactionOptions` but with a single field `SubmitForSettlement bool`? That's the only field listed in the Braintree Java SDK.

transaction.go
```diff
@@ -97,6 +97,30 @@ type TransactionRequest struct {
...
+	ServiceFeeAmount   *Decimal                  `xml:"service-fee-amount,attr,omitempty"`
+	RiskData           *RiskDataRequest          `xml:"risk-data,omitempty"`
+	Descriptor         *Descriptor               `xml:"descriptor,omitempty"`
+	CustomFields       customfields.CustomFields `xml:"custom-fields,omitempty"`
```
leighmcculloch: Comparing this struct to the [TransactionCloneRequest](https://github.com/braintree/braintree_java/blob/acc7c9ada19e09f6181dd501657fc7f4f1b7afcb/src/main/java/com/braintreegateway/TransactionCloneRequest.java) in the Braintree Java SDK, many of these options shouldn't be included in the clone request. The only fields this struct should have are:

* `Amount *Decimal`
* `Channel string`
* `Options TransactionCloneOptions`

dsmcfarl:
@leighmcculloch No problem, thanks for maintaining these bindings. Let me know if the latest changes I pushed don't address your suggestions.

leighmcculloch:
The changes in this PR are in `v0.14.0` which has been released.

PR #187, opened 2017-09-15
Fix flaky subscription billing day of month tests

leighmcculloch:
What
===
Allow subscription tests that set a billing day of month to pass if
their status is active and not pending after creation.

Why
===
The tests use a fixed billing day of month of the 15th of the month. If
the subscriptions are created on the 15th, the subscription will go
active immediately rather than be returned in a pending state. This was
causing the tests to fail on the 15th of each month.


leighmcculloch:
The changes in this PR are in `v0.14.0` which has been released.

PR #189, opened 2017-09-29
Move pointers test helpers to testhelpers package

leighmcculloch:
What
===
Move test helper functions for creating int and bool type pointers to
the `testhelpers` package.

Why
===
The `testhelpers` package is where other helper functions exist already,
and it's better if we have them in one place.


leighmcculloch:
Released in `v0.15.0`.

Issue #179, opened 2017-07-19
Customer update with VerifyCard: false returns "Gateway Rejected: cvv"


ghost:
Hi, I've been using braintree-go and I've come into an issue where I'm trying to update a customer's credit card billing address without using a payment method nonce, and I'm getting the `"Gateway Rejected: cvv"` error response.

I have the `VerifyCard` field set to `false` in `CreditCardOptions`, which I believe should help me avoid getting this error response.

Anyways, my code is at the bottom.  Let me know if anyone else has seen this same issue or if I'm doing anything incorrectly.

```go
c, err := s.api.Customer().Update(
	&braintree.Customer{
		Id: req.UserToken,
		CreditCard: &braintree.CreditCard{
			Options: &braintree.CreditCardOptions{
				UpdateExistingToken: req.PaymentMethodToken,
				VerifyCard:          false,
			},
			CardholderName: req.CardholderName,
			BillingAddress: &braintree.Address{
				StreetAddress:     req.BillingAddress.StreetAddress,
				ExtendedAddress:   req.BillingAddress.ExtendedAddress,
				PostalCode:        req.BillingAddress.PostalCode,
				CountryCodeAlpha2: req.BillingAddress.CountryCodeAlpha2,
			},
		},
	},
)
```

leighmcculloch:
I tried to replicate the above on my sandbox account, but in every case I did not get the `Gateway Rejected: cvv` error that you did. It's possible that there's a setting configured in your account that's different to mine though that might cause verification to kick in. The `VerifyCard` parameter is only sent to the API if it is `true`. Setting it to `false` is the same as not setting it.

esrobinson:
Yeah, I think this happens when the account is set up to verify card by default.  In that case, to add an address or update a card without providing the CVV, you have to pass `VerifyCard` as `false` to the API.

leighmcculloch:
@ghost @esrobinson: Does #190 address the issue you're seeing?

esrobinson:
Yeah, I think that takes care of it.  Thanks!

leighmcculloch:
Fixed is released in `v0.15.0`.

PR #190, opened 2017-09-29
Allow card verification to be turned off

leighmcculloch:
What
===
Change the type of the `VerifyCard` option from `bool` to `*bool`.

Why
===
For merchants who have payment method verification required in their
control panel, they are unable to set the value of `VerifyCard` to
`false` and disable verification for specific payment methods like they
can with other Braintree SDKs.

This happens because the `VerifyCard` field of `CreditCardOptions` and
`PaymentMethodRequestOptions` is tagged as `omitempty`, because we only
want to send the value to the gateway if the merchant has explicitly set
a value. Combined with the type of `bool`, this means if a merchant sets
the value as `false`, we don't send it to the gateway.

Making this field a `*bool` and `omitempty` will mean the field is not
sent if it is `nil` (not set), but will be sent if it is given the value
of `false` or `true.

This issue was raised in #179.

Note
===
There is additional marshaling code added to make the VerifyCard field
marshal correctly on versions of go prior to go1.8.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -23,6 +23,10 @@ When using Braintree Go in a production environment, we recommend that you conti
 
 #### Sandbox settings
 
+In your sandbox account go to `Settings > Processing` and enable the following
+
+  1. `Card Verification`
+
 In your sandbox account go to `Settings > Processing > CVV` and enable the following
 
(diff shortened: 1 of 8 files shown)


leighmcculloch:
Released in `v0.15.0`.

PR #194, opened 2017-11-03
Add context.Context to exported functions that make outgoing remote calls

### LINKED ISSUE #156: support context.Context in all remote APIs

leighmcculloch:
What
===
Add context to all exported functions that make an outgoing remote call. Insert as the first parameter in every function in the package that make an outgoing remote call a `context.Context`.

Why
===
To give the caller control over canceling remote calls.

Resolves #156
Supersedes #158 

@lionelbarrow @jmhodges


leighmcculloch:
@ulisesflynn @ccahoon @rkulla @RichardKnop @brmasson @michaljemala @tamalsaha @dsmcfarl @icza @atishpatel @lquach @sadlil @bastiankoetsier: You folks have all contributed in the last few months. I'm looking for feedback on this change, not for detailed code review but rather high-level confirmation that this would mesh well with your usage of the package.

bastiankoetsier:
@leighmcculloch LGTM, every outgoing request should be cancelable 👍

atishpatel:
This is a pretty big code breaking change, but I can't think of a way to implement it without breaking people's code unless we made new functions. I guess people who aren't vendoring their libraries will just have to deal with it. ¯\_(ツ)_/¯ 

You updated the README, but there is another example in `doc.go` that should have context. Otherwise, LGTM.

dsmcfarl:
@leighmcculloch LGTM. Shouldn't be a problem to integrate into my code. You did the hard work!

leighmcculloch:
@atishpatel I agree, it's a big breaking change. Because of the best practice in the go docs that context should be the first parameter the only other option is to double the number of functions in the package and that's make the package API more daunting I think.

PR #198, opened 2017-11-17
Add PurchaseOrderNumber to Transaction and TransactionRequest.

dsmcfarl:
This should be added because adding a purchase order number to commercial card transactions lowers clearing rates by allowing Level II processing.


leighmcculloch:
Thanks for adding this @dsmcfarl! ❤️

PR #209, opened 2018-01-31
added Addresses to Customer and Update to AddressGateway

brandoneprice31:
Allow users to get all of their addresses and update them.

### DIFF
--- a/address_gateway.go
+++ b/address_gateway.go
@@ -39,3 +39,21 @@ func (g *AddressGateway) Delete(ctx context.Context, customerId, addrId string) 
 	}
 	return &invalidResponseError{resp}
 }
+
+// Updates an address for the specified id and customer id.
+func (g *AddressGateway) Update(ctx context.Context, customerId, addrId string, a *Address) (*Address, error) {
+	// Copy address so that field sanitation won't affect original
+	var cp Address = *a
+	cp.CustomerId = ""
+	cp.XMLName = xml.Name{Local: "address"}
+
+	resp, err := g.execute(ctx, "PUT", "customers/"+customerId+"/addresses/"+addrId, &cp)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


address_gateway.go
```diff
@@ -39,3 +39,21 @@ func (g *AddressGateway) Delete(ctx context.Context, customerId, addrId string)
 	}
 	return &invalidResponseError{resp}
 }
+
+// Updates an address for the specified id and customer id.
+func (g *AddressGateway) Update(ctx context.Context, customerId, addrId string, a *Address) (*Address, error) {
+	// Copy address so that field sanitation won't affect original
+	var cp Address = *a
+	cp.CustomerId = ""
```
leighmcculloch: Instead of sanitizing the address fields, I think we need to employ the same pattern here that we've moved to in other gateways, where there is an `AddressRequest` for creates and updates, and an `Address` for finds. I  can see this follows the same pattern right now as the create though.

customer.go
```diff
@@ -22,6 +22,7 @@ type Customer struct {
 	AndroidPayCards    *AndroidPayCards          `xml:"android-pay-cards,omitempty"`
 	ApplePayCards      *ApplePayCards            `xml:"apple-pay-cards,omitempty"`
 	PaymentMethodNonce string                    `xml:"payment-method-nonce,omitempty"`
+	Addresses          *Addresses                `xml:"addresses,omitempty"`
```
leighmcculloch: We can't really add this here because `Customer` is used for both requests and responses and it isn't valid to pass addresses when creating a customer. That has to be done separately through the address gateway.

leighmcculloch:
@brandoneprice31 We're just missing a test now for the `AddressGateway` `Update` function you added. If you can add that, this can be merged. 🎉

PR #213, opened 2018-02-27
Add skip fraud checking to transaction options

ulisesflynn:


transaction_integration_test.go
```diff
@@ -657,6 +657,7 @@ func TestAllTransactionFields(t *testing.T) {
 			StoreInVault:                     true,
 			AddBillingAddressToPaymentMethod: true,
 			StoreShippingAddressInVault:      true,
+			SkipAdvancedFraudChecking:        true,
```
leighmcculloch: This line is causing the test to fail because there are other expectations on the response returned that are not present when you set this flag.

ulisesflynn:
I've fixed up the tests, thank you for the quick feedback!

leighmcculloch:
@ulisesflynn Thanks for fixing up the tests!

Hey if you're going to be at GopherCon this year, let me know. I'm `@leigh` on Gophers Slack.

PR #218, opened 2018-03-26
Add SubscriptionDetails to Transaction struct

ducnt114:


ducnt114:
I have replaced `BillingPeriod` with existing `Subscription` struct and updated test case for new field.

leighmcculloch:
Three tests are failing, but two of them are also failing on master so we can ignore those. The one that failed had this error message:
>--- FAIL: TestAllTransactionFields (4.03s)
>	transaction_integration_test.go:888: expected tx2.RiskData.ID not to be empty

I'm re-running the build to see if it was temporal.

leighmcculloch:
The rerun of the tests showed that test pass, so looks good!

leighmcculloch:
@ducnt114 I just took a look at the other Braintree SDKs and I realized that other SDKs actually organize this structure the same way you had originally, but with a different name. e.g. The ruby SDK has a `SubscriptionDetails` with the two billing period fields. Looking at the developer docs those are the only two fields that get returned. Sorry for the misdirection asking you to make it like the Java SDK.

I've changed it back to how you had it with the new struct instead, and renamed it to `SubscriptionDetails`.

I also added a test for when the transaction is not part of a subscription and added custom unmarshalling logic so that the `SubscriptionDetails` is `nil` on the `Transaction` when there is no subscription.

PR #220, opened 2018-04-23
Add StatusHistory to Subscription

Shugyousha:
Having this status change information is useful for us and this lets us
get it.

Original code by alexander@smallpdf.com.


status_history.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// StatusEvent contains information about what and when something
+// changed about the subscription.
+type StatusEvent struct {
+	Timestamp          time.Time `xml:"timestamp"`
+	Balance            *Decimal  `xml:"balance"`
+	Price              *Decimal  `xml:"price"`
+	Status             string    `xml:"status"`
+	SubscriptionSource string    `xml:"subscription-source"`
```
leighmcculloch: Could you also include the `currency-iso-code`, `user`, and `plan-id` fields?

status_history.go
```diff
@@ -0,0 +1,21 @@
...
+
+// StatusHistory contains information about the last 50 timestamps where
+// something changed about the subscription.
+type StatusHistory struct {
+	StatusEvents []StatusEvent `xml:"status-event"`
+}
+
+// StatusEvent contains information about what and when something
+// changed about the subscription.
+type StatusEvent struct {
```
leighmcculloch: Lets call these types `SubscriptionStatusHistory` and `SubscriptionStatusEvent` because there are also status events for transactions that haven't been added to the this SDK yet, and they have different fields so it would be difficult to share this type with transactions.

subscription.go
```diff
@@ -43,6 +43,7 @@ type Subscription struct {
 	TrialPeriod             bool                 `xml:"trial-period"`
 	Transactions            *Transactions        `xml:"transactions"`
 	Options                 *SubscriptionOptions `xml:"options"`
+	StatusHistory           *StatusHistory       `xml:"status-history"`
```
leighmcculloch: Is it possible to omit the `StatusHistory` type and simply make this `[]*StatusEvent` with a tag `xml:"status-history>status-event"`?

status_history.go
```diff
@@ -0,0 +1,21 @@
...
+	Timestamp          time.Time `xml:"timestamp"`
+	Balance            *Decimal  `xml:"balance"`
+	Price              *Decimal  `xml:"price"`
+	Status             string    `xml:"status"`
```
leighmcculloch: Could you make this a type, like `TransactionStatus` is a type?

https://github.com/lionelbarrow/braintree-go/blob/5ae1956a21cedc7ac4a4fe0b5234d2f72c3f0733/transaction.go#L10-L27

subscription.go
Shugyousha: I don't think having slices with pointers is a good idea. Why not just []StatusEvent instead? That way there is no need to check a pointer for nil every time you iterate over the slice

status_history.go
Shugyousha: "plan-id" is not part of the status events but of the Transaction itself:

// TODO: not all transaction fields are implemented yet, here are the missing fields (add on demand)
//
// <transaction>
//   <voice-referral-number nil="true"></voice-referral-number>
//   <status-history type="array">
//     <status-event>
//       <timestamp type="datetime">2013-10-07T17:26:14Z</timestamp>
//       <status>authorized</status>
//       <amount>7.00</amount>
//       <user>eaigner</user>
//       <transaction-source>Recurring</transaction-source>
//     </status-event>
//     <status-event>
//       <timestamp type="datetime">2013-10-07T17:26:14Z</timestamp>
//       <status>submitted_for_settlement</status>
//       <amount>7.00</amount>
//       <user>eaigner</user>
//       <transaction-source>Recurring</transaction-source>
//     </status-event>
//     <status-event>
//       <timestamp type="datetime">2013-10-08T07:06:38Z</timestamp>
//       <status>settled</status>
//       <amount>7.00</amount>
//       <user nil="true"></user>
//       <transaction-source></transaction-source>
//     </status-event>
//   </status-history>
//   <plan-id>bronze</plan-id>
//   <subscription-id>jqsydb</subscription-id>
//   <subscription>
//     <billing-period-end-date type="date">2013-11-06</billing-period-end-date>
//     <billing-period-start-date type="date">2013-10-07</billing-period-start-date>
//   </subscription>
//   <add-ons type="array"/>
//   <discounts type="array"/>
//   <descriptor>
//     <name nil="true"></name>
//     <phone nil="true"></phone>
//   </descriptor>
//   <recurring type="boolean">true</recurring>
// </transaction>

from transaction.go

The rest of the fields I will add.

Shugyousha: Argh, never mind. I looked at the Transaction history, not the Subscription history. I will add the plan-id field as well.

status_history.go
Shugyousha: Ok, I will change the filename accordingly.

Shugyousha:
I don't think the Travis CI failures are due to my changes.

Let me know if you want me to change something else.

subscription.go
```diff
@@ -17,35 +17,36 @@ const (
...
+	PaymentMethodToken      string                    `xml:"payment-method-token"`
+	PlanId                  string                    `xml:"plan-id"`
+	Price                   *Decimal                  `xml:"price"`
+	Status                  SubscriptionStatus        `xml:"status"`
+	TrialDuration           string                    `xml:"trial-duration"`
+	TrialDurationUnit       string                    `xml:"trial-duration-unit"`
+	TrialPeriod             bool                      `xml:"trial-period"`
+	Transactions            *Transactions             `xml:"transactions"`
+	Options                 *SubscriptionOptions      `xml:"options"`
+	StatusHistory           []SubscriptionStatusEvent `xml:"status-history>status-event"`
```
leighmcculloch: Could you call this field `StatusEvents` since it is a slice of StatusEvents?

leighmcculloch: I think this reference should still be a pointer for consistency with the other types in the SDK. I agree with your point that the pointer is unnecessary, but at this point the consistency is worth it I think.

leighmcculloch:
If you merge master into your branch it should go green again. The tests on master have been fixed.

Shugyousha:
I did the requested changes and added some tests.

Travis CI is still failing but the issue seems to be an unrelated test ( TestMerchantAccountTransaction ) that is failing.

subscription_integration_test.go
```diff
@@ -47,6 +47,25 @@ func TestSubscriptionSimple(t *testing.T) {
 	if sub.Id == "" {
 		t.Fatal("invalid subscription id")
 	}
+	if len(sub.StatusEvents) != 1 {
+		t.Fatalf("expected one status event, got %d", len(sub.StatusEvents))
+	}
+	wantBalance := NewDecimal(0, 2)
+	wantPrice := NewDecimal(1000, 2)
+	for _, event := range sub.StatusEvents {
```
leighmcculloch: Because there's only one status event and you check that a few lines above, we can drop the for loop here and do `sub.StatusEvents[0]`.

PR #221, opened 2018-04-30
Add RetryCharge function

Shugyousha:
We use this functionality in our code so we would like to see this function added.

Note that the test needs some manual setup in order to be useful.

Let me know if you want me to change something.


Shugyousha:
The CI system does not seem to like my use of the assert package. Do you want me to remove its use or to add it to the CI system somehow?

subscription_integration_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/lionelbarrow/braintree-go/testhelpers"
+	"github.com/stretchr/testify/assert"
```
leighmcculloch: I like testify personally too, but the braintree-go package has no external dependencies which makes it much easier for consumers of the package to use it. Can we drop this and replace it with standard Go tests?

Once dependency management becomes simplified in go with vgo, I think we can move to a test framework.

subscription_integration_test.go
```diff
@@ -1012,3 +1013,26 @@ func TestSubscriptionTransactions(t *testing.T) {
...
+
+// In order to run this test you have to create a test subscription
+// that has a status of "Past Due". You can get such a subscription
+// by creating one with a trial period of one day and a price of 2000
+// USD (this amount will fail in the sandbox environment). After the
+// trial expires, the subscription will have the desired "Past Due"
+// status. Then you can try to charge an amount lower than 2000 which
+// then should work when we retry the charge.
+func TestSubscriptionRetryCharge(t *testing.T) {
+	t.Skip("Needs manual setup")
```
leighmcculloch: I'm going to investigate if we can avoid manual setup for this test. Thanks for including a test!!!

subscription_gateway.go
```diff
@@ -53,3 +56,33 @@ func (g *SubscriptionGateway) Cancel(ctx context.Context, subId string) (*Subscr
...
+// be > 0.
+func (g *SubscriptionGateway) RetryCharge(ctx context.Context, subId string, amount Decimal) error {
+	txInput := &struct {
+		XMLName        xml.Name
+		Amount         Decimal            `xml:"amount"`
+		Options        TransactionOptions `xml:"options"`
+		SubscriptionID string             `xml:"subscription-id"`
+		Type           string             `xml:"type"`
+	}{
+		XMLName: xml.Name{Local: "transaction"},
```
leighmcculloch: I think you should be able to move the `"transaction"` into a tag on the `xml.Name` field like:

```
XMLName xml.Name `xml:"transaction"`
```

subscription_gateway.go
```diff
@@ -53,3 +56,33 @@ func (g *SubscriptionGateway) Cancel(ctx context.Context, subId string) (*Subscr
 	}
 	return nil, &invalidResponseError{resp}
 }
+
+// RetryCharge retries to charge for a subscription. The amount has to
+// be > 0.
+func (g *SubscriptionGateway) RetryCharge(ctx context.Context, subId string, amount Decimal) error {
```
leighmcculloch: We should give the user more control over the transaction that gets created out of the retry. For example in the Braintree Java SDK there are four functions:

https://github.com/braintree/braintree_java/blob/8fa6a13/src/main/java/com/braintreegateway/SubscriptionGateway.java#L108-L132

```java
public Result<Transaction> retryCharge(String subscriptionId) { ... }
public Result<Transaction> retryCharge(String subscriptionId, BigDecimal amount) { ... }
public Result<Transaction> retryCharge(String subscriptionId, Boolean submitForSettlement) { ... }
public Result<Transaction> retryCharge(String subscriptionId, BigDecimal amount, Boolean submitForSettlement) { ... }
```

I think we should probably create an exported struct `SubscriptionTransactionRequest` like what the Braintree Java SDK has, and have that be the only data the user needs to provide. e.g. https://github.com/braintree/braintree_java/blob/8fa6a13/src/main/java/com/braintreegateway/SubscriptionTransactionRequest.java

subscription_integration_test.go
Shugyousha: Where I work we couldn't find a way. If you find one please let us know! :)

leighmcculloch:
Thanks @Shugyousha! I made a couple changes to hide static fields in the XML request from the public API. I couldn't find a way as you had said to run a successful RetryCharge, so I changed your test so that it could run all the time. It attempts a retry and expects it to fail saying the status is not PastDue. I think this is better than having a manual test we never run. It at least validates the other fields we're sending are correct.

subscription_integration_test.go
```diff
@@ -1013,41 +1013,84 @@ func TestSubscriptionTransactions(t *testing.T) {
...
+		Options: &SubscriptionTransactionOptionsRequest{
 			SubmitForSettlement: true,
 		},
-		Type: "sale",
+	})
+	if err != nil {
+		if err, ok := err.(*BraintreeError); ok {
+			t.Log(err)
+			fieldErrors := err.All()
+			if g, w := len(fieldErrors), 1; g != w {
```
Shugyousha: That seems unnecesserily complicated. Why not just

if len(fieldErrors) != 1 {

?

subscription_integration_test.go
```diff
@@ -1013,41 +1013,84 @@ func TestSubscriptionTransactions(t *testing.T) {
...
+				t.Fatalf("got %d field errors, want %d", g, w)
+			}
+			if g, w := fieldErrors[0].Code, "91531"; g != w {
```
Shugyousha: Same here:

if fieldErrors[0].Code != "91531" {

is easier to read and more common.

subscription_integration_test.go
```diff
@@ -1013,41 +1013,84 @@ func TestSubscriptionTransactions(t *testing.T) {
...
+				t.Errorf("got error code %q, want %q", g, w)
+			}
+			if g, w := fieldErrors[0].Message, "Subscription status must be Past Due in order to retry."; g != w {
```
Shugyousha: Same as above

subscription_integration_test.go
```diff
@@ -1013,41 +1013,84 @@ func TestSubscriptionTransactions(t *testing.T) {
...
+				t.Errorf("got error message %q, want %q", g, w)
+			}
+		} else {
```
Shugyousha: You could get rid of one level of indentation by checking for

if err, ok := err.(*BraintreeError); !ok {
  t.Fatal(err)
}

above

subscription_integration_test.go
```diff
@@ -1013,41 +1013,84 @@ func TestSubscriptionTransactions(t *testing.T) {
...
+			t.Fatal(err)
+		}
+	} else {
+		t.Errorf("Retry charge did not error, want error indicating Subscription status must be Past Due in order to retry.")
```
Shugyousha: Same here. You could just do

if err == nil {
  t.Fatalf("Retry charge did not error, want error indicating Subscription status must be Past Due in order to retry.")
}

further up.

Shugyousha:
Other than that it looks good to me.

leighmcculloch:
@Shugyousha Thanks for the feedback, addressed!

Shugyousha:
Looks good to me!

leighmcculloch:
Thanks @Shugyousha !

PR #222, opened 2018-05-05
Rewrite validation errors handling

leighmcculloch:
Rewrite the parsing of validation errors, so that all objects and fields returned in an error response are available. Motivation for doing this is because #214 adds line items but the existing error code can't support dynamic error responses that contains errors for parameter arrays.

In the process of rewriting this I found several bugs with how validation errors were parsed, and they're fixed by this change. Examples:
1. You no longer need to call `.For("Transaction").For("Base").For("Base")` to get to the base errors on a Transaction, instead you call `.For("Transaction").For("Base")`.
2. There is no longer an inconsistency between the casing of object names and attributes, as all are normalized to their matching struct fields, which is title camel case. e.g. `CreditCard`, `DiscountAmount`.

Since so much was changing, I also normalized some of the names of structs with the names used in the official Braintree SDKs, so it's easier to understand how to use the this SDK by reading the standard Braintree developer docs. e.g. `FieldError` => `ValidationError`.

Because of the numerous changes this change is a breaking change, but only for consumers who were using the `For` and `On` functions to find specific errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 0.19.0 (Unreleased)
+
+BACKWARDS INCOMPATIBILITES:
+
+* Rewrite error handling and interfaces. See #222 for details.
+
 ## 0.18.0 (March 2nd, 2018)
 
 BACKWARDS INCOMPATIBILITES:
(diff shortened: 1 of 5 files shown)


leighmcculloch:
@lionelbarrow Got any feels about this?

Issue #197, opened 2017-11-07
Subscriptions.Search


munknex:
Would be nice to add this function:

[https://developers.braintreepayments.com/reference/request/subscription/search/python](https://developers.braintreepayments.com/reference/request/subscription/search/python)

leighmcculloch:
@munknex would you like to open a PR adding it? The search objects already exist because transaction search is already implemented. It just needs expanding to Subscription. Related files you could look at for more context:

https://github.com/lionelbarrow/braintree-go/blob/a246878e7d43d6b5dc38ec985ab7b6312cf4df46/search.go#L8-L42

https://github.com/lionelbarrow/braintree-go/blob/a246878e7d43d6b5dc38ec985ab7b6312cf4df46/transaction_gateway.go#L156-L167

https://github.com/lionelbarrow/braintree-go/blob/a246878e7d43d6b5dc38ec985ab7b6312cf4df46/transaction.go#L214-L220

leighmcculloch:
Resolved in #229.

PR #229, opened 2018-05-28
added Search functionality for subscriptions

tengis617:
**Reference:** #197  

I would like to implement the Search functionality for subscriptions. 
As a starting point, i added the SearchIDs function to subscriptions

### DIFF
--- a/subscription_integration_test.go
+++ b/subscription_integration_test.go
@@ -1272,6 +1272,7 @@ func TestSubscriptionSearchPagination(t *testing.T) {
 		if err != nil {
 			t.Fatal(err)
 		}
+		expectedIDs[sub.Id] = true
 	}
 
 	t.Logf("expectedIDs = %v", expectedIDs)
--- a/subscription_integration_test.go
+++ b/subscription_integration_test.go
@@ -1173,13 +1173,10 @@ func TestSubscriptionSearchIDs(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if len(result.IDs) < 2 {
-		t.Errorf("expected length of result.IDs to be greater than 2, got %v", len(result.IDs))
-	}
 	if !testhelpers.Contains(result.IDs, sub1.Id) {
 		t.Errorf("expected result.IDs to include %v", sub1.Id)
 	}
-	if !testhelpers.Contains(result.IDs, sub1.Id) {
+	if !testhelpers.Contains(result.IDs, sub2.Id) {
 		t.Errorf("expected result.IDs to include %v", sub2.Id)
 	}
 	if testhelpers.Contains(result.IDs, sub3.Id) {
@@ -1228,9 +1225,6 @@ func TestSubscriptionSearch(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	if len(result.Subscriptions) < 2 {
-		t.Errorf("expected length of result.IDs to be greater than 2, got %v", len(result.TotalIDs))
-	}
 	if result.CurrentPageNumber != 1 {
 		t.Errorf("expected page number to be 1, got %v", result.CurrentPageNumber)
 	}
@@ -1238,10 +1232,10 @@ func TestSubscriptionSearch(t *testing.T) {
 		t.Errorf("expected subscription ids to contain %v", sub1.Id)
 	}
 	if !testhelpers.Contains(result.TotalIDs, sub2.Id) {
-		t.Errorf("expected subscription ids to contain %v", sub1.Id)
+		t.Errorf("expected subscription ids to contain %v", sub2.Id)
 	}
 	if testhelpers.Contains(result.TotalIDs, sub3.Id) {
-		t.Errorf("expected subscription ids to not contain %v", sub1.Id)
+		t.Errorf("expected subscription ids to not contain %v", sub3.Id)
 	}
 }
 
@@ -1293,7 +1287,7 @@ func TestSubscriptionSearchPagination(t *testing.T) {
 	t.Logf("result.PageSize = %v", result.PageSize)
 
 	if result.TotalItems < subscriptionCount {
-		t.Fatalf("result.TotalItems = %v, want it to be more than %v", result.TotalItems, subscriptionCount)
+		t.Errorf("result.TotalItems = %v, want it to be more than %v", result.TotalItems, subscriptionCount)
 	}
 
 	for {
--- a/subscription_integration_test.go
+++ b/subscription_integration_test.go
@@ -1156,11 +1156,18 @@ func TestSubscriptionSearchIDs(t *testing.T) {
 		PaymentMethodToken: paymentMethod.GetToken(),
 		PlanId:             "test_plan",
 	})
+	sub3, err := g.Create(ctx, &SubscriptionRequest{
+		PaymentMethodToken: paymentMethod.GetToken(),
+		PlanId:             "test_plan_2",
+	})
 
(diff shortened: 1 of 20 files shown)


testhelpers/slice.go
```diff
@@ -0,0 +1,12 @@
...
+// Contains checks whether a slice of strings contains
+// a given string
+func Contains(list []string, s string) bool {
+	for _, item := range list {
+		if item == s {
+			return true
+		}
+	}
+	return false
+}
```
leighmcculloch: Could you add a test for this function?

subscription_integration_test.go
```diff
@@ -1129,3 +1129,50 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+	f.Min = time.Now().AddDate(0, 0, -1)
+	result, err := g.SearchIDs(ctx, query)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(result.IDs) < 2 {
+		t.Errorf("expected length of result.IDs to be greater than 2, got %v", len(result.IDs))
+	}
+	if !contains(result.IDs, sub1.Id) {
```
leighmcculloch: `contains` => `testhelpers.Contains`

subscription_integration_test.go
```diff
@@ -1129,3 +1129,50 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+		t.Errorf("expected result.IDs to include %v", sub1.Id)
+	}
+	if !testhelpers.contains(result.IDs, sub1.Id) {
```
leighmcculloch: `testhelpers.contains` => `testhelpers.Contains`

subscription_integration_test.go
```diff
@@ -1129,3 +1129,50 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+	g := testGateway.Subscription()
+
+	sub1, err := g.Create(ctx, &SubscriptionRequest{
+		PaymentMethodToken: paymentMethod.GetToken(),
+		PlanId:             "test_plan",
+	})
+	sub2, err := g.Create(ctx, &SubscriptionRequest{
+		PaymentMethodToken: paymentMethod.GetToken(),
+		PlanId:             "test_plan",
+	})
```
leighmcculloch: Can you add a third subscription that has some property, that will cause it not to be included in the test? To do this you'll need to search on more than `created-at`. This is to ensure that the query is actually being used to filter and we're not just getting back all recent subscriptions.

tengis617: @leighmcculloch would it be better if i add more tests to check different edge cases? I think just having a single test might actually not be the best idea. Will definitely add the test case you mentioned.

leighmcculloch: Having a single test with the additional cases within is 👍 .

leighmcculloch:
A test failure was fixed on `master` in #232. Could you merge `master` into your branch?

subscription_integration_test.go
```diff
@@ -1129,3 +1129,190 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	if len(result.IDs) < 2 {
+		t.Errorf("expected length of result.IDs to be greater than 2, got %v", len(result.IDs))
+	}
+	if !testhelpers.Contains(result.IDs, sub1.Id) {
+		t.Errorf("expected result.IDs to include %v", sub1.Id)
+	}
+	if !testhelpers.Contains(result.IDs, sub1.Id) {
```
leighmcculloch: This line has a typo on it, it is `sub1.Id`, but should be `sub2.Id`.

subscription_integration_test.go
```diff
@@ -1129,3 +1129,190 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+	f2.Is = "test_plan"
+
+	result, err := g.SearchIDs(ctx, query)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(result.IDs) < 2 {
+		t.Errorf("expected length of result.IDs to be greater than 2, got %v", len(result.IDs))
+	}
```
leighmcculloch: I think you can delete these three lines. The lines below will error appropriately if only one was returned.

subscription_integration_test.go
```diff
@@ -1129,3 +1129,190 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+	f2.Is = "test_plan"
+
+	result, err := g.Search(ctx, query)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if len(result.Subscriptions) < 2 {
+		t.Errorf("expected length of result.IDs to be greater than 2, got %v", len(result.TotalIDs))
+	}
```
leighmcculloch: You can drop these three lines, per my similar comment above.

subscription_integration_test.go
```diff
@@ -1129,3 +1129,190 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+		t.Errorf("expected page number to be 1, got %v", result.CurrentPageNumber)
+	}
+	if !testhelpers.Contains(result.TotalIDs, sub1.Id) {
+		t.Errorf("expected subscription ids to contain %v", sub1.Id)
+	}
+	if !testhelpers.Contains(result.TotalIDs, sub2.Id) {
+		t.Errorf("expected subscription ids to contain %v", sub1.Id)
+	}
+	if testhelpers.Contains(result.TotalIDs, sub3.Id) {
+		t.Errorf("expected subscription ids to not contain %v", sub1.Id)
```
leighmcculloch: Typo, should be: `sub3`.

subscription_integration_test.go
```diff
@@ -1129,3 +1129,190 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+		t.Errorf("expected length of result.IDs to be greater than 2, got %v", len(result.TotalIDs))
+	}
+	if result.CurrentPageNumber != 1 {
+		t.Errorf("expected page number to be 1, got %v", result.CurrentPageNumber)
+	}
+	if !testhelpers.Contains(result.TotalIDs, sub1.Id) {
+		t.Errorf("expected subscription ids to contain %v", sub1.Id)
+	}
+	if !testhelpers.Contains(result.TotalIDs, sub2.Id) {
+		t.Errorf("expected subscription ids to contain %v", sub1.Id)
```
leighmcculloch: Typo, should be: `sub2`.

subscription_integration_test.go
```diff
@@ -1129,3 +1129,190 @@ func TestSubscriptionRetryCharge(t *testing.T) {
...
+	result, err := g.Search(ctx, query)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("result.TotalItems = %v", result.TotalItems)
+	t.Logf("result.TotalIDs = %v", result.TotalIDs)
+	t.Logf("result.PageSize = %v", result.PageSize)
+
+	if result.TotalItems < subscriptionCount {
+		t.Fatalf("result.TotalItems = %v, want it to be more than %v", result.TotalItems, subscriptionCount)
```
leighmcculloch: I think this could be `t.Errorf`. We don't need to fail here do we?

tengis617:
Sorry guys, i was on vacation for a few days. I just fixed all the tests according to the reviews. Hope we can close this soon!

leighmcculloch:
I expect to be able to merge this in Friday. I also plan to cut a new release them.

PR #231, opened 2018-05-29
Added Dispute structs and ability to parse them from WebhookNotificat…

### LINKED ISSUE #228: Support Dispute object in webhooks
1) Need to define respective structs and support unmarshalling of `braintree.WebhookNotification.Subject` to a `Dispute`.
2) Currently no need to implement API handles like [this](https://developers.braintreepayments.com/reference/request/dispute/accept/ruby).

If this feature is OK for you - can send respective PR.

vania-pooh:
…ion (fixes #228)


leighmcculloch:
@vania-pooh A test failure was fixed on `master` in #232. Could you merge `master` into your branch?

vania-pooh:
@leighmcculloch done.

vania-pooh:
@leighmcculloch now tests seem to pass.

webhook_notification_test.go
```diff
@@ -249,3 +249,99 @@ func TestWebhookParseDisbursementException(t *testing.T) {
...
+
+	notification, err := webhookGateway.Parse(signature, payload)
+
+	if err != nil {
+		t.Fatal(err)
+	} else if notification.Kind != DisputeOpenedWebhook {
+		t.Fatal("Incorrect Notification kind, expected dispute got", notification.Kind)
+	} else if notification.Dispute() == nil {
+		t.Fatal("Notification should have a dispute")
+	} else if notification.Dispute().Kind != DisputeKindChargeback {
```
leighmcculloch: All the assertions past this point don't need to use `Fatalf`. They could use `Errorf` and be standalone `if` statements so that if multiple of them are wrong it will report all of the fields that have failed assertions instead of just the first one.

vania-pooh:
@leighmcculloch let's me see whether it's complicated to implement the gateway.

leighmcculloch:
Looking at the line item gateway I just added in acaf667871d84e1a955317e5f5003fd78596a004 should provide a good example of what required.

vania-pooh:
@leighmcculloch see my latest changes. I implemented the entire gateway however I'm stuck with integration tests. A `Dispute` seems to be a read-only object, so I can't create it in test with something like `dispute.Create()` and then use respective `disputeId`. Any idea how to test it?

leighmcculloch:
@vania-pooh It's worth having a read through the [Disputes](https://developers.braintreepayments.com/guides/disputes/overview) documentation. The most relevant page will be the [Testing and Go Live](https://developers.braintreepayments.com/guides/disputes/testing-go-live/java) page that has some examples for how to test a dispute in sandbox. Basically you create a transaction with a set card number, set amount, and then it will be immediately disputed, and you can submit evidence to simulate the dispute being won or lost.

dispute.go
```diff
@@ -0,0 +1,65 @@
...
+	DisputeReasonTransactionAmountDiffers      = "transaction_amount_differs"
+
+	// Dispute statuses
+	DisputeStatusAccepted = "accepted"
+	DisputeStatusDisputed = "disputed"
+	DisputeStatusExpired  = "expired"
+	DisputeStatusOpen     = "open"
+	DisputeStatusLost     = "lost"
+	DisputeStatusWon      = "won"
+)
```
leighmcculloch: For each of these could you add a type, so that they are not just strings, and make the corresponding fields on the structs those types? Example:

https://github.com/lionelbarrow/braintree-go/blob/273799669d13c0589383b83d45239c5b5819528a/transaction.go#L10-L17

dispute.go
```diff
@@ -0,0 +1,65 @@
...
+
+type Dispute struct {
+	XMLName         string             `xml:"dispute"`
+	AmountDisputed  *Decimal           `xml:"amount-disputed"`
+	AmountWon       *Decimal           `xml:"amount-won"`
+	CaseNumber      string             `xml:"case-number"`
+	CreatedAt       *time.Time         `xml:"created-at"`
+	CurrencyISOCode string             `xml:"currency-iso-code"`
+	Evidence        []*DisputeEvidence `xml:"evidence>evidence"`
+
```
leighmcculloch: What's the motivation for splitting the two sets of fields?

dispute_gateway.go
```diff
@@ -0,0 +1,137 @@
...
+	*Braintree
+}
+
+type FileEvidenceRequest struct {
+	XMLName    xml.Name `xml:"evidence"`
+	DocumentId string   `xml:"document-id"`
+	Category   string   `xml:"category,omitempty"`
+}
+
+type TextEvidenceRequest struct {
```
leighmcculloch: Would it make these structs clearer if they were prefixed with `Dispute`? Looking at the other Braintree SDKs, they don't seem to have it prefixed, but I think it would make it clearer in the godocs for this SDK, since they would be grouped. What do you think?

dispute_gateway.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var v DisputeSearchResult
+	err = xml.Unmarshal(resp.Body, &v)
+	if err != nil {
+		return nil, err
+	}
+	return &v, err
+}
```
leighmcculloch: I think we'll need a couple more `Search` functions. `SearchIDs`, `Search` and `SearchNext`, like here:

https://github.com/lionelbarrow/braintree-go/blob/273799669d13c0589383b83d45239c5b5819528a/transaction_gateway.go#L156-L158

https://github.com/lionelbarrow/braintree-go/blob/273799669d13c0589383b83d45239c5b5819528a/transaction_gateway.go#L181-L183

https://github.com/lionelbarrow/braintree-go/blob/273799669d13c0589383b83d45239c5b5819528a/transaction_gateway.go#L212-L215

But I think that's a pretty large piece of work and I think it's worth dropping the `Search` function for now, and lets do that in a separate PR once this is merged. The main goal for doing some of these other functions now is to make sure the modeling of `Dispute`, `DisputeEvidence`, `DisputeTransactionDetails`, and we should get that from the other functions you're adding in the gateway.

vania-pooh: @leighmcculloch did not find such implementations even in official Java client. Do such API handles exist (can't find the way to verify)?

dispute.go
vania-pooh: Agreed.

dispute.go
vania-pooh: Don't get the question. Just specify how would you implement this.

dispute_gateway.go
vania-pooh: No problem. Will rename the structs.

dispute.go
vania-pooh: `Evidence` was declared like this: https://github.com/lionelbarrow/braintree-go/blob/master/subscription.go#L48

leighmcculloch: I was meaning what's the reason for the empty line between the fields? It's fine though.

vania-pooh: @leighmcculloch ah, ok. That was done accidentally. Will fix.

vania-pooh:
@leighmcculloch added gateway and integration tests. However it throws a strange 403 Forbidden for `addFileEvidence` method. Here I need your knowledge to understand why this happens. There is nothing about it in Braintree docs.

dispute_evidence.go
```diff
@@ -0,0 +1,12 @@
+package braintree
+
+import "time"
+
+type DisputeEvidence struct {
+	XMLName           string     `xml:"evidence"`
+	Comment           string     `xml:"comment"`
+	CreatedAt         *time.Time `xml:"created-at"`
+	ID                string     `xml:"id"`
+	SentToProcessorAt string     `xml:"sent-to-processor-at"`
```
leighmcculloch: I want to make this field and the other date fields being added an object, maybe `*time.Time`, but it's not straight forward because the object is only a date and not a full date time. 🤔

vania-pooh:
@leighmcculloch removed useless methods and now all tests seem to pass. I would merge this as is. We could add modifications related to dates and document in next PRs. Otherwise we have too many changes in one PR.

leighmcculloch:
I expect to be able to merge this on Friday. I also plan to cut a new release them.

vania-pooh:
@leighmcculloch ok, agreed.

leighmcculloch:
@vania-pooh Just took a look and when you remove some things you force pushed the branch and lost the other changes I made. Preferably lets avoid force pushing.

leighmcculloch:
I've just merged back in my changes and the original history before the force push, preserving the changes in the force push.

leighmcculloch:
@vania-pooh There's no need to keep rebasing. We can push changes as single commits onto the branch so that neither of us lose changes.

vania-pooh:
@leighmcculloch already done.

leighmcculloch:
@vania-pooh Unfortunately you just blew away my previous changes again.

vania-pooh:
@leighmcculloch omg. Seeing for the first time that anybody except me is pushing to PR branch.

leighmcculloch:
Hang tight, I'll remerge my changes back in. 😃

leighmcculloch:
Okay fixed back up. From here on, lets just push new commits to the branch. 😄

vania-pooh:
@leighmcculloch ok.

leighmcculloch:
Thanks @vania-pooh ! This was a big effort. I created new issues to track the things that we decided to pull out of this PR. Let me know if you're interested in picking up either. The new issues are:

#239 – Support for Dispute Search
#240 – Support for Dispute File Evidence

PR #236, opened 2018-06-01
Add Current billing cycle field to Discount

Shugyousha:
(The first commit is just a drive-by cleanup)

This field is useful if you want to determine if a Discount is still ongoing or not.

With the plans in the default environment this is difficult to test (because they only apply the discount after a trial/at the end of the month) but I verified that the field is being picked up on our installation.


leighmcculloch:
Thanks @Shugyousha!

PR #243, opened 2018-06-24
Changed Subscription.Update() method to support updating ID (fixes #242)

### LINKED ISSUE #242: Subscription.Update() method does not allow to update subscription ID
Just compare. In Java client it works like the following:
```
SubscriptionRequest request = new SubscriptionRequest()
  .id("new_id")
  .paymentMethodToken("new_payment_method_token")
  .price(new BigDecimal("14.00"))
  .planId("new_plan")
  .merchantAccountId("new_merchant_account");

Result<Subscription> result = gateway.subscription().update(
  "a_subscription_id",
  request
);
```
In this Go client the value from `SubscriptionRequest` is always used. This prevents user from updating the ID. So I think a correct signature should be `Update(ctx context.Context, id string, sub *SubscriptionRequest)`.

vania-pooh:


vania-pooh:
@leighmcculloch changed in `TestSubscriptionAllFieldsWithBillingDayOfMonth`. Let's see whether the tests will pass. :)

subscription_integration_test.go
```diff
@@ -195,8 +199,7 @@ func TestSubscriptionAllFieldsWithBillingDayOfMonth(t *testing.T) {
 	}
 
 	// Update
-	sub2, err := g.Update(ctx, &SubscriptionRequest{
-		Id:     sub1.Id,
+	sub2, err := g.Update(ctx, sub1.Id, &SubscriptionRequest{
 		PlanId: "test_plan_2",
```
leighmcculloch: Can you add an expectation for this update that checks that the ID of the audition returned starts the same? There's already an expectation to check if the ID is not an empty string so that just needs changing to check if it requires the original ID.

leighmcculloch: Where is that? I see this: https://github.com/lionelbarrow/braintree-go/pull/243/files#diff-fe0b52236b5a897ed31d1dac80a712b6R283

leighmcculloch: Ahh I see it now, here https://github.com/vania-pooh/braintree-go/blob/945a8238b432278e9d47aab6267fe04614083ac3/subscription_integration_test.go#L216.

leighmcculloch:
Thanks @vania-pooh !!!! ❤️

PR #244, opened 2018-06-26
Add method ExpiringBetween to CreditCardGateway

ducnt114:


credit_card_gateway.go
```diff
@@ -57,3 +62,30 @@ func (g *CreditCardGateway) Delete(ctx context.Context, card *CreditCard) error
 	}
 	return &invalidResponseError{resp}
 }
+
+//  ExpiringBetween Find list of credit card that expire between the specified dates
+func (g *CreditCardGateway) ExpiringBetween(ctx context.Context, fromDate, toDate time.Time) ([]*CreditCard, error) {
+	path := fmt.Sprintf("/payment_methods/all/expiring?start=%s&end=%s",
+		fromDate.Format("012006"),
+		toDate.Format("012006"))
```
leighmcculloch: According to the braintree/braintree_java SDK we should send these dates in UTC.
Ref: [braintree/braintree_java](https://github.com/braintree/braintree_java/blob/2def3f6/src/main/java/com/braintreegateway/CreditCardGateway.java#L211)

credit_card_gateway.go
```diff
@@ -57,3 +62,30 @@ func (g *CreditCardGateway) Delete(ctx context.Context, card *CreditCard) error
...
+	resp, err := g.execute(ctx, "POST", path, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	type searchResult struct {
+		XMLName           xml.Name      `xml:"payment-methods"`
+		CurrentPageNumber int64         `xml:"current-page-number"`
+		PageSize          int64         `xml:"page-size"`
```
leighmcculloch: How do we use the paging fields of this struct with the function? Or does this function only support the first page? If it only supports the first page, can we use the same pattern that was added in 55b62f92375c92ad5903deb2b4dc71065ccd3171 and using both the `/payment_methods/all/expiring_ids` and `/payment_methods/all/expiring` endpoints?  
Ref: [braintree/braintree_java](https://github.com/braintree/braintree_java/blob/2def3f6dcf3fbbfcb3c74d67716aadfa3f0658bd/src/main/java/com/braintreegateway/CreditCardGateway.java#L190-L207)

I think it'd also be worth adding the `Expired()` function too alongside `ExpiringBetween` since it should follow the same pattern.  
Ref: [braintree/braintree_java](https://github.com/braintree/braintree_java/blob/2def3f6dcf3fbbfcb3c74d67716aadfa3f0658bd/src/main/java/com/braintreegateway/CreditCardGateway.java#L164-L180)

ducnt114:
Thanks @leighmcculloch , I have added method Expired() and some tests in `credit_card_integration_test.go`, also update the last commit to follow same pattern as Java SDK.

ducnt114:
@leighmcculloch can you review this pull request? I have updated and added some tests.

leighmcculloch:
@ducnt114 I'm reviewing this now, I'm going to make some changes directly to your PR, and then merge it. Please don't force push it from here on :).

leighmcculloch:
@ducnt114 The Braintree developer documentation only lists `ExpiringBetween`, not `Expired`, and it's technically not possible to test `Expired` properly, especially with our the braintree-go's sandbox account because the account has more than 10,000 expired credit cards and only the first 10k seem to get returned. Both of these things is leading me to think we should leave the `Expired` function out, since `ExpiringBetween` is basically the same thing with a window. I'll remove it from the PR when merging, but the code is still here in this PR for us to revisit if we want. Let me know what you think.

leighmcculloch:
This is in a state now where it's consistent with other functions that support paging through larger sets of data, but before I merge this I'm going to see if I can introduce a better way to page. I found it difficult to write the code using the existing patterns.

leighmcculloch:
I've opened PR #246 which changes the pagination pattern a little for existing pagination implementations. If we were to use the same pattern here, it would look like this change b5474fc.

ducnt114:
@leighmcculloch I agree with you as we leave method `Expired()` out and keep `ExpiringBetween()`.
When I added 2 methods, I realize that we need to paging result but with current coding pattern, we can get only 1 page.

PR #245, opened 2018-07-05
added createdAt and updatedAt to subscription

tengis617:
added `created-at` to subscription

### DIFF
--- a/subscription_integration_test.go
+++ b/subscription_integration_test.go
@@ -197,9 +197,8 @@ func TestSubscriptionAllFieldsWithBillingDayOfMonth(t *testing.T) {
 	if x := sub1.Descriptor.URL; x != "example.com" {
 		t.Fatalf("got descriptor url %#v, want example.com", x)
 	}
-	year, month, day := time.Now().Date()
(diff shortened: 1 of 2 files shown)


subscription_integration_test.go
```diff
@@ -197,6 +197,10 @@ func TestSubscriptionAllFieldsWithBillingDayOfMonth(t *testing.T) {
 	if x := sub1.Descriptor.URL; x != "example.com" {
 		t.Fatalf("got descriptor url %#v, want example.com", x)
 	}
+	year, month, day := time.Now().Date()
+	if x := sub1.CreatedAt; x.Year() != day || x.Month() != month || x.Day != day {
+		t.Fatalf("got %v %v %v , want %v %v %v", x.Year(), x.Month(), x.Day(), year, month, day)
+	}
```
leighmcculloch: This test could be flaky. It's best if it can just check it's not nil.

tengis617: @leighmcculloch i see. Let me fix that real quick

tengis617:
@leighmcculloch im adding `updated-at` as well

leighmcculloch:
Thanks!!!!

PR #247, opened 2018-07-06
Add SearchPage to TransactionGateway and SubscriptionGateway

leighmcculloch:
What
===
Add `SearchPage` to `TransactionGatewag` and `SubscriptionGateway` that
can be used to page through search results returned by `SearchIDs`.
Deprecate other search functions, `Search` and `SearchNext`.

Why
===
The existing pagination pattern of using the `Search` and `SearchNext`
functions are not simple to program with, and they limit access to the
search space and assume the developer will want to search sequentially
through the pages.

The complexity with using the functions primarily comes from the fact
that the developer gets paged results from two different functions. This
means the developer cannot put those functions merely in a for loop and
iterate. They must call one first, deal with those results, then loop on
the second function. It is less intuitive to write code this way, and
much easier to think about if we get the IDs first, then just iterate
over them in pages.

### DIFF
--- a/search.go
+++ b/search.go
@@ -21,8 +21,9 @@ func (s *SearchQuery) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
 }
 
 type SearchResult struct {
-	PageSize int
-	IDs      []string
+	PageSize  int
+	PageCount int
+	IDs       []string
 }
 
 type TextField struct {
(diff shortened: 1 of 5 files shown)


leighmcculloch:
@dsmcfarl @rkulla @ilayaraja89 You commented on the previous issues #206 and #112 relating to this. Feedback appreciated on this attempt at simplifying pagination.

PR #248, opened 2018-07-13
Adding StoreInVaultOnSuccess in TransactionOptions

jvalecillos:
Thank you @lionelbarrow for all your work on this excellent client library for Braintree API. In this PR I am adding an additional option for controlling when the tokens are stored in Braintree's Vault:

- [x] Adding `StoreInVaultOnSuccess` field for `TransactionOptions`. From the [docs](https://developers.braintreepayments.com/reference/request/transaction/sale/php#options.store_in_vault_on_success):
> The option that determines whether the payment method associated with the successful transaction should be stored in the Vault.

- [x] Adding also tests for `StoreInVault`.

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -177,6 +177,7 @@ type Transactions struct {
 type TransactionOptions struct {
 	SubmitForSettlement              bool                             `xml:"submit-for-settlement,omitempty"`
 	StoreInVault                     bool                             `xml:"store-in-vault,omitempty"`
+	StoreInVaultOnSuccess            bool                             `xml:"store-in-vault-on-success,omitempty"`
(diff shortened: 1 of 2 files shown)


jvalecillos:
I'm working on the tests. The actual behaviour in the API is that the tokens (and also the customers) are stored in the Vault even of the payment is refused. But it seems like the transaction is null (nil) and there is just an error.

jvalecillos:
Tests passing now :heavy_check_mark:

leighmcculloch:
Thanks @jvalecillos! Thanks for submitting the change and including tests, as well as including tests for the other flag as well! Great use of table-driven tests too.

PR #249, opened 2018-07-17
Added PaymentInstrumentType constants for Transaction object

vania-pooh:


vania-pooh:
@leighmcculloch it's better to have constants than using literals everywhere in the code. :)

transaction.go
```diff
@@ -35,56 +35,68 @@ const (
 	TransactionSourceMerchant       TransactionSource = "merchant"
 )
 
+type TransactionPaymentInstrumentType string
```
leighmcculloch: I think this could be named `PaymentInstrumentType`. That's what the enum is called in the Braintree .NET SDK.

https://github.com/braintree/braintree_dotnet/blob/882a6ea/src/Braintree/Transaction.cs#L110-L130

vania-pooh:
@leighmcculloch done.

transaction.go
```diff
@@ -35,6 +35,18 @@ const (
 	TransactionSourceMerchant       TransactionSource = "merchant"
 )
 
+type PaymentInstrumentType string
+
+const (
+	TransactionPaymentInstrumentTypeAndroidPayCard   PaymentInstrumentType = "android_pay_card"
+	TransactionPaymentInstrumentTypeApplePayCard     PaymentInstrumentType = "apple_pay_card"
+	TransactionPaymentInstrumentTypeCreditCard       PaymentInstrumentType = "credit_card"
```
leighmcculloch: Can you also update the names of the constants to match the too?

vania-pooh:
Done.

leighmcculloch:
Thanks @vania-pooh!!!

PR #257, opened 2018-07-23
Better customer search with paging (fixes #256)

### LINKED ISSUE #256: Fetch all customers
In my code I have a periodic background process that needs to somehow fetch `all` customers via Braintree API and then do some processing for every customer (mainly related to manually handling trial expiration). Right now `CustomerGateway` has `Search(ctx context.Context, query *SearchQuery)` method to get a list of customers.

**My question:** is a single call to this method with an empty search query sufficient to get a list of all customers?

My intuition says - something is missing in the client as I'm seeing `CurrentPageNumber`, `PageSize` and `TotalItems` fields in the `CustomerSearchResult` structure but no way to use these fields. So how would I fetch all the customers? If something is missing I could send respective PR.

vania-pooh:
Closes #256

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -75,9 +75,10 @@ type CustomerRequest struct {
 }
 
 type CustomerSearchResult struct {
-	XMLName           string      `xml:"customers"`
-	CurrentPageNumber int         `xml:"current-page-number"`
-	PageSize          int         `xml:"page-size"`
-	TotalItems        int         `xml:"total-items"`
-	Customers         []*Customer `xml:"customer"`
+	TotalItems int
+	TotalIDs   []string
+
+	CurrentPageNumber int
+	PageSize          int
+	Customers         []*Customer
 }
--- a/customer_gateway.go
+++ b/customer_gateway.go
@@ -84,27 +84,7 @@ func (g *CustomerGateway) Search(ctx context.Context, query *SearchQuery) (*Cust
 		return nil, err
 	}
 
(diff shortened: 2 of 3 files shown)


customer.go
```diff
@@ -75,9 +75,10 @@ type CustomerRequest struct {
...
-	CurrentPageNumber int         `xml:"current-page-number"`
-	PageSize          int         `xml:"page-size"`
-	TotalItems        int         `xml:"total-items"`
-	Customers         []*Customer `xml:"customer"`
+	XMLName           string
+	CurrentPageNumber int
+	TotalItems        int
+	TotalIDs          []string
+	PageSize          int
+	Customers         []*Customer
```
leighmcculloch: Could you organize these fields so they're in the same order as the equivalent `Transaction`, `Subscription` and `CreditCard` types? The consistency is helpful, and it separating the fields relating to a single page, and the fields relating to all of the results makes it easier to reason about.

https://github.com/lionelbarrow/braintree-go/blob/d4e31c11b0086e116fee1d2b9631d634226d245e/transaction.go#L255-L262

https://github.com/lionelbarrow/braintree-go/blob/d4e31c11b0086e116fee1d2b9631d634226d245e/subscription.go#L120-L127

vania-pooh: Ok, will fix today.

leighmcculloch:
Thanks @vania-pooh !

PR #265, opened 2018-11-04
Refactoring client token generate function

jvalecillos:
This refactoring `Generate` function from `ClientTokenGateway` to align it with the [official libraries](https://developers.braintreepayments.com/reference/request/client-token/generate/php).

This change support all documented parameters but also fix certain issues for some users. In particular, when the user wants to use `ApplePay` or `GooglePlay` and has more than one merchant account (E.g. sub accounts per country), the client token generation defaults to the main account if no `merchantAccountId` is given. This causes issues on the client SDK (browser or mobile) if, for example, `ApplePay` or `GooglePlay` is not enabled in the country for the default account.


client_token_gateway.go
```diff
@@ -7,23 +7,23 @@ import (
...
 	*Braintree
 }
 
 // Generate generates a new client token.
-func (g *ClientTokenGateway) Generate(ctx context.Context) (string, error) {
-	return g.generate(ctx, &ClientTokenRequest{
-		Version: clientTokenVersion,
-	})
-}
+func (g *ClientTokenGateway) Generate(ctx context.Context, req *ClientTokenRequest) (string, error) {
```
leighmcculloch: I think in the case of client token generation it's worth leaving the existing functions as they are and adding a new function `GenerateWithRequest` that exposes the full variation of options. The function hasn't exposed the additional fields and options until now so we can make a guess that existing users of the package don't need the additional functionality and we should avoid inconveniencing them by breaking the API.

client_token_integration_test.go
```diff
@@ -8,39 +8,69 @@ import (
...
+		{
+			name: "request with customer id and merchant id and options",
+			req: &ClientTokenRequest{
+				CustomerID:        customer.Id,
+				MerchantAccountID: testMerchantAccountId,
+				Options: &ClientTokenRequestOptions{
+					FailOnDuplicatePaymentMethod: true,
+					MakeDefault:                  true,
+					VerifyCard:                   true,
+				},
```
leighmcculloch: 🎉 These table-driven tests are a fantastic improvement, thank you!

examples/subscription/server.go
```diff
@@ -37,7 +37,7 @@ func main() {
 
 	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
 		ctx := context.Background()
-		clientToken, err := bt.ClientToken().Generate(ctx)
+		clientToken, err := bt.ClientToken().Generate(ctx, nil)
```
leighmcculloch: 👏 Thanks for updating the examples.

client_token_integration_test.go
```diff
@@ -8,39 +8,69 @@ import (
...
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			g := testGateway.ClientToken()
+			token, err := g.Generate(context.TODO(), tt.req)
+			if (err != nil) != tt.wantError {
+				t.Errorf("ClientTokenGateway.Generate() error = %v, wantErr = %v", err, tt.wantError)
+				return
+			}
+			if !tt.wantErr && len(token) == 0 {
```
leighmcculloch: This line has a typo on it that is what's causing the CI builds to fail. `wantError` is the name of the field on `tt`.

Aside to that though, I think `err` is a common name for an identifier containing an error, so think it's worth renaming the field to `wantErr` to match this instead.

client_token.go
```diff
@@ -1,11 +1,21 @@
 package braintree
 
+// ClientTokenRequest represents the parameters for the client token request
 type ClientTokenRequest struct {
-	XMLName    string `xml:"client-token"`
-	CustomerID string `xml:"customerId,omitempty"`
-	Version    int    `xml:"version"`
+	XMLName           string                     `xml:"client-token"`
+	CustomerID        string                     `xml:"customerId,omitempty"`
+	MerchantAccountID string                     `xml:"merchantAccountId,omitempty"`
```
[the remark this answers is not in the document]
jvalecillos: Well, I changed it but I cannot guarantee it is the required tag name for the XML since even the official library it is not consistent about that. Please take a look here:

https://github.com/braintree/braintree_java/blob/2def3f6dcf3fbbfcb3c74d67716aadfa3f0658bd/src/main/java/com/braintreegateway/SubscriptionTransactionRequest.java#L36
https://github.com/braintree/braintree_java/blob/2def3f6dcf3fbbfcb3c74d67716aadfa3f0658bd/src/main/java/com/braintreegateway/PaymentMethodRequest.java#L137
https://github.com/braintree/braintree_java/blob/2def3f6dcf3fbbfcb3c74d67716aadfa3f0658bd/src/main/java/com/braintreegateway/ClientTokenOptionsRequest.java#L53

It would be easier if the documentation included the schema (WSDL?) for the XML instead of referring to the SDKs variable names.

Do you have any insight about this?.

jvalecillos:
I had to tests in a row failing about the amount of the transaction, but only in one version of go at the time and randomly
``` bash
go test -parallel 15 -tags='unit integration' ./...
--- FAIL: TestTransactionSearchNext (36.22s)
    transaction_integration_test.go:287: Amount must be greater than zero.
```
[1](https://travis-ci.com/braintree-go/braintree-go/jobs/156320611) and [2](https://travis-ci.com/braintree-go/braintree-go/jobs/156323626)

I think it should be completely unrelated with the PR itself.

leighmcculloch:
You can ignore the `TestTransactionSearchNext` failures, I'm guessing the test setup must not be ideal. I've created issue #266 to follow up on that separately.

client_token.go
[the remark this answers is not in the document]
leighmcculloch: The different SDKs (Ruby, Java, .NET, etc) all use different casing for the fields and they all seem to be interpreted appropriately. The Java SDK uses camelCase as you pointed out in many places. The Ruby SDK appears to exclusively use kebab case as can be seen in it's [XML generator](https://github.com/braintree/braintree_ruby/blob/2df252641403a86e2212d9c9b3bd8849569c3e5d/lib/braintree/xml/generator.rb#L101). Both camelCase and kebab-case can be used for each field

jvalecillos:
Rebased from `master` to have the latest fixes from the tests.

leighmcculloch:
Whoops, one thing I forgot was that versions of Go prior to 1.8 have a bug where `*bool` are not serialized properly. We should manually write out the `MarshalXML` function for the like has been done here for the other `*bool` fields. e.g.: 

https://github.com/braintree-go/braintree-go/blob/abe29b19ee6b20d619a49772d73b1e8427c6ea47/go_fixes_bug_5452_xml_omitempty.go#L17-L59

I can pick this up in the next few days, or if you'd like to contribute that change too, you're welcome.

PR #268, opened 2018-11-06
Custom xml marshalling for ClientTokenRequestOptions

jvalecillos:
Following up #265 and fixing issue regarding XML serialization of Boolean pointer for `ClientTokenRequestOptions`.

### DIFF
--- a/go_fixes_bug_5452_xml_omitempty.go
+++ b/go_fixes_bug_5452_xml_omitempty.go
@@ -14,6 +14,7 @@ import "encoding/xml"
 // To serialize the bool false value when it is set on `VerifyCard`, we must manually control
 // if it is serialized or not.
 
+// MarshalXML custom serialization for CreditCardOptions.
 func (cco *CreditCardOptions) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
 	if cco.VerifyCard == nil {
 		type excludeVerifyCard struct {
@@ -58,6 +59,7 @@ func (cco *CreditCardOptions) MarshalXML(e *xml.Encoder, start xml.StartElement)
 	}
 }
 
+// MarshalXML custom serialization for PaymentMethodRequestOptions.
 func (pmo *PaymentMethodRequestOptions) MarshalXML(e *xml.Encoder, start xml.StartElement) error {
 	if pmo.VerifyCard == nil {
 		type excludeVerifyCard struct {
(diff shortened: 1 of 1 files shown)


go_fixes_bug_5452_xml_omitempty.go
```diff
@@ -24,72 +28,61 @@ func (cco *CreditCardOptions) MarshalXML(e *xml.Encoder, start xml.StartElement)
...
+	}
+	type includeVerifyCard struct {
+		VerifyCard                    *bool  `xml:"verify-card"`
+		VenmoSDKSession               string `xml:"venmo-sdk-session,omitempty"`
+		MakeDefault                   bool   `xml:"make-default,omitempty"`
+		FailOnDuplicatePaymentMethod  bool   `xml:"fail-on-duplicate-payment-method,omitempty"`
+		VerificationMerchantAccountId string `xml:"verification-merchant-account-id,omitempty"`
+		UpdateExistingToken           string `xml:"update-existing-token,omitempty"`
 	}
+	return e.EncodeElement(includeVerifyCard(*cco), start)
```
leighmcculloch: Unfortunately the ability to convert from one type to another that has the same fields and differs only with it's tags was introduced after Go 1.7, so we can't use it in the SDK yet.

leighmcculloch:
@jvalecillos I added a series a test for marshaling `ClientToken` in #269 that demonstrates the serialization issue with Go 1.7. I'll merge it to master shortly, and then you can merge master into your PR to confirm if your change fixes it.

go_fixes_bug_5452_xml_omitempty.go
jvalecillos: It is fixed now. I was not aware that the casting was relatively new feature of the language.

jvalecillos:
@leighmcculloch You read my mind. I even created a test locally and I noticed that the `MarshalXML` functions failed when trying to serialize a null pointer for the type. That's why my second commit accounts for it adding an additional case to avoid the panic on `if cco.VerifyCard == nil`. You maybe want to add that case in the tests.

leighmcculloch:
Ah good idea with the `nil` case. I've committed the tests and included a `nil` scenario. Are you able to reduce this change down to just adding `MarshalXML` for `ClientTokenRequestOptions` and break-out the `nil` changes for the others to a separate PR?

jvalecillos:
@leighmcculloch Yeah, sure. It's done. I'm creating a separate PR based on this one with the two additional commits.

go_fixes_bug_5452_xml_omitempty.go
```diff
@@ -14,6 +14,7 @@ import "encoding/xml"
 // To serialize the bool false value when it is set on `VerifyCard`, we must manually control
 // if it is serialized or not.
 
+// MarshalXML custom serialization for CreditCardOptions.
```
leighmcculloch: Thanks for adding these comments!

PR #270, opened 2018-11-08
Amount to decimal

edwardbrowncross:
Use Decimal type for amount rather than string.

This is made possible by the `EncodeValues` function, which is expected by the go-querystring library:
https://github.com/google/go-querystring/blob/master/query/encode.go


edwardbrowncross:
Wrong repo. Sorry!

leighmcculloch:
@edwardbrowncross No worries! Look forward to seeing it when you open the change here.

PR #271, opened 2018-11-08
Some refactoring and more tests around custom marshaling functions

jvalecillos:
# What
Handling null pointer serialization case and added unit tests around it.

# Why
The tests will represent a form of documentation about the expected scenarios for the request options. In this case, there are 3 possible values to send to the API: nothing (null), true or false.

This PR is based on #268


leighmcculloch:
Ah, I see the issue. The issue is you're calling `MarshalXML` on the object directly, which we should never need to do. If you'd like to serialize without indenting you can use `xml.Marshal` and that function like `xml.MarshalIndent`. Both of those functions in the `xml` package guard against calling `MarshalXML` on objects if the object is `nil`.

Do you have a specific need where you need to be able to call `MarshalXML` directly on the object?

jvalecillos:
@leighmcculloch Actually not at all. I think is a corner case that will never happen in newer versions of Go. Considering that, and given that my changes were removed we could or close the PR or keep it for the sake of unit test coverage.

leighmcculloch:
Yup it's not something that should happen with Go1.7 either. I ran the tests locally. Yup we should keep these tests. These tests are more thorough than the existing tests and so are a good addition.

Issue #272, opened 2018-11-08
CreditCard `product_id` not included on CreditCard type?


krak3n:
Hi, thank you for producing this library, I was wondering why the `product_id` field is not included on the CreditCard type? https://developers.braintreepayments.com/reference/response/credit-card/python#product_id

Thanks,

Chris

leighmcculloch:
@krak3n I don't think there's any particular reason it hasn't been included other than nobody has added it yet. Would you be up for opening a PR to add this field?

krak3n:
@leighmcculloch certainly :+1: gives me something to do on the train home :train:

leighmcculloch:
The PR for this #274 has been merged. Thanks @krak3n !

PR #274, opened 2018-11-09
Feature: add new ProductId field to CreditCard type

krak3n:
See #272 for context.

Adds new field `ProductId` to the `CreditCard` model.

Couldn't see any tests that maybe  affect this, let me know if I need to add any :+1:

### DIFF
--- a/credit_card.go
+++ b/credit_card.go
@@ -39,7 +39,7 @@ type CreditCard struct {
 	UniqueNumberIdentifier    string             `xml:"unique-number-identifier,omitempty"`
 	BillingAddress            *Address           `xml:"billing-address,omitempty"`
 	Subscriptions             *Subscriptions     `xml:"subscriptions,omitempty"`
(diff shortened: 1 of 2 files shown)


credit_card.go
```diff
@@ -39,6 +39,7 @@ type CreditCard struct {
 	UniqueNumberIdentifier    string             `xml:"unique-number-identifier,omitempty"`
 	BillingAddress            *Address           `xml:"billing-address,omitempty"`
 	Subscriptions             *Subscriptions     `xml:"subscriptions,omitempty"`
+	ProductId                 string             `xml:"product_id,omitempty"`
```
leighmcculloch: `ProductID` to stick with Go's naming conventions.
`product-id` for the XML to keep the serialization consistent. Both works, but we've been trying to use `-` consistently.

krak3n:
Oh cool thanks :+1:

PR #275, opened 2018-11-09
Add `AuthorizationExpiresAt` to `Transaction`

emdotcarter:
What
----
Add `AuthorizationExpiresAt` to `Transaction`

Why
---
This field is now supported by the other SDKs, providing a datetime that
the authorization will expire (if it is still in authorized state)o

https://developers.braintreepayments.com/reference/response/transaction/ruby#authorization_expires_at

### DIFF
--- a/transaction_integration_test.go
+++ b/transaction_integration_test.go
@@ -990,6 +990,9 @@ func TestAllTransactionFields(t *testing.T) {
 	if tx2.SubscriptionDetails != nil {
 		t.Fatalf("expected Subscription to be not nil, but got %#v", tx2.SubscriptionDetails)
 	}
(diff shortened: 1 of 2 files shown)


transaction_integration_test.go
```diff
@@ -990,6 +990,9 @@ func TestAllTransactionFields(t *testing.T) {
 	if tx2.SubscriptionDetails != nil {
 		t.Fatalf("expected Subscription to be not nil, but got %#v", tx2.SubscriptionDetails)
 	}
+	if tx2.AuthorizationExpiresAt == nil {
```
leighmcculloch: Could we test that the value isn't only not `nil` but that it also has a value in a range of time that would be appropriate?

PR #277, opened 2018-11-20
Add a RefundWithRequest method to TransactionGateway

### LINKED ISSUE #276: (*TransactionGateway).Refund does not support specifying an order ID
It looks like the Ruby version of the Braintree SDK allows a order ID to be specified as part of creating a refund: https://developers.braintreepayments.com/reference/request/transaction/refund/ruby#order_id

braintree-go does not allow this.

I imagine that this is only a matter of adding an order ID to the `TransactionRequest`. If so, would you like me to open a pull request to do this? And would you like to create a new variant of `Refund` with a different argument list, or modify the existing one in a backward-incompatible way?

aaronlehmann:
The current `Refund` method cannot take additional parameters like
`OrderId`. This new method allows those parameters to be passed as part of
a request structure.

Fixes #276

### DIFF
--- a/transaction.go
+++ b/transaction.go
@@ -130,6 +130,12 @@ type TransactionRequest struct {
 	LineItems           TransactionLineItemRequests `xml:"line-items,omitempty"`
 }
 
+type TransactionRefundRequest struct {
+	XMLName string   `xml:"transaction"`
+	Amount  *Decimal `xml:"amount"`
+	OrderID string   `xml:"order-id,omitempty"`
+}
+
 func (t *Transaction) UnmarshalXML(d *xml.Decoder, start xml.StartElement) error {
 	type typeWithNoFunctions Transaction
 	if err := d.DecodeElement((*typeWithNoFunctions)(t), &start); err != nil {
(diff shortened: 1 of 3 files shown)


transaction.go
```diff
@@ -130,6 +130,12 @@ type TransactionRequest struct {
 	LineItems           TransactionLineItemRequests `xml:"line-items,omitempty"`
 }
 
+type TransactionRefundRequest struct {
+	XMLName string   `xml:"transaction"`
+	Amount  *Decimal `xml:"amount"`
+	OrderId string   `xml:"order-id,omitempty"`
```
leighmcculloch: Can we have this parameter `OrderID`? There's a mix of casings in this package but trying to make new ones consistent with Go standards.

transaction_gateway.go
```diff
@@ -140,6 +140,22 @@ func (g *TransactionGateway) Refund(ctx context.Context, id string, amount ...*D
 	return nil, &invalidResponseError{resp}
 }
 
+// RefundWithRequest is similar to Refund, but allows more parameters such as
+// OrderId to be included as part of the refund request.
```
leighmcculloch: `OrderID`

aaronlehmann:
Thanks. I've amended the commit to change `OrderId` to `OrderID`.

Issue #278, opened 2019-01-30
Unit amount invalid format ".66"


midoNuminix:
Hello, 

Thanks for the library. When representing a  value `NewDecimal(66,2)` it returns the result as `.66`. As a result, braintree returns invalid format for line item unit amount as it expects the value to be `0.66`. A fix will be doing the following: 

```
// MarshalText outputs a decimal representation of the scaled number
func (d *Decimal) MarshalText() (text []byte, err error) {
	b := new(bytes.Buffer)
	if d.Scale <= 0 {
		b.WriteString(strconv.FormatInt(d.Unscaled, 10))
		b.WriteString(strings.Repeat("0", -d.Scale))
	} else {
		str := strconv.FormatInt(d.Unscaled, 10)

		if len(str) < d.Scale {

			str = strings.Repeat("0", d.Scale) + str
		}

		z := str[:len(str)-d.Scale]
		if z == "" {
			z = "0" // To avoid doing strings.Repeat("0", 1) + str
		}
		b.WriteString(z)
		b.WriteString(".")
		b.WriteString(str[len(str)-d.Scale:])
	}
	return b.Bytes(), nil
}

```

leighmcculloch:
Thanks for raising this behavior @midoNuminix. I've made a change #280 to address the serialization inconsistency of `Decimal`.

leighmcculloch:
This was fixed in #280.

PR #280, opened 2019-03-03
Serialize decimals to string consistently

leighmcculloch:
What
===
Change the number of zeros added to the beginning of a decimal that will
only render non-zero digits below a decimal point so that there is
always one zero before the decimal point.

Why
===
The decimal serialization to a string appears to have a couple of
oddities.

`NewDecimal(55, 2)` serialized to `".55"`.
`NewDecimal(55, 3)` serialized to `"00.055"`.

When calculating the number of zeros to add to the beginning of a number
if the scale is larger than the number, it wasn't taking into account if
the length was equal to the scale (e.g. 0.55 where there are two digits
provided and a scale of 2), and the other situation above. The number of
zeros being added was the scale and not the number of zeros missing to
consistently generate a string leading with `0.`.

This shouldn't have caused any issues because the resulting string is still
a valid decimal according to many decimal parsers, but as pointed out by
@midoNuminix in #278 some fields appear to not like the values.

### DIFF
--- a/decimal.go
+++ b/decimal.go
@@ -28,8 +28,8 @@ func (d *Decimal) MarshalText() (text []byte, err error) {
 		b.WriteString(strings.Repeat("0", -d.Scale))
 	} else {
 		str := strconv.FormatInt(d.Unscaled, 10)
-		if len(str) < d.Scale {
-			str = strings.Repeat("0", d.Scale) + str
+		if len(str) <= d.Scale {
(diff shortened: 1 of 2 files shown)


leighmcculloch:
@midoNuminix does this fix work for you?

midoNuminix:
@leighmcculloch , yup it does. Thanks  :100:

PR #283, opened 2019-04-05
Added BillingAddress field to PaymentMethodRequest

lindroth:
Why
===
The BrainTree api supports providing billing address when creating a payment method.
It is also the only way to set it on a payment method as the api for credit card requires PCI SAQ D compliance.
https://developers.braintreepayments.com/reference/request/payment-method/create/ruby#billing_address

### DIFF
--- a/address.go
+++ b/address.go
@@ -26,19 +26,19 @@ type Address struct {
 }
 
 type AddressRequest struct {
-	XMLName            xml.Name `xml:"address"`
-	FirstName          string   `xml:"first-name,omitempty"`
-	LastName           string   `xml:"last-name,omitempty"`
-	Company            string   `xml:"company,omitempty"`
-	StreetAddress      string   `xml:"street-address,omitempty"`
-	ExtendedAddress    string   `xml:"extended-address,omitempty"`
-	Locality           string   `xml:"locality,omitempty"`
-	Region             string   `xml:"region,omitempty"`
-	PostalCode         string   `xml:"postal-code,omitempty"`
-	CountryCodeAlpha2  string   `xml:"country-code-alpha2,omitempty"`
-	CountryCodeAlpha3  string   `xml:"country-code-alpha3,omitempty"`
-	CountryCodeNumeric string   `xml:"country-code-numeric,omitempty"`
-	CountryName        string   `xml:"country-name,omitempty"`
+	XMLName            xml.Name
+	FirstName          string `xml:"first-name,omitempty"`
+	LastName           string `xml:"last-name,omitempty"`
+	Company            string `xml:"company,omitempty"`
+	StreetAddress      string `xml:"street-address,omitempty"`
+	ExtendedAddress    string `xml:"extended-address,omitempty"`
+	Locality           string `xml:"locality,omitempty"`
+	Region             string `xml:"region,omitempty"`
+	PostalCode         string `xml:"postal-code,omitempty"`
+	CountryCodeAlpha2  string `xml:"country-code-alpha2,omitempty"`
+	CountryCodeAlpha3  string `xml:"country-code-alpha3,omitempty"`
+	CountryCodeNumeric string `xml:"country-code-numeric,omitempty"`
+	CountryName        string `xml:"country-name,omitempty"`
 }
 
 type Addresses struct {
--- a/payment_method_gateway.go
+++ b/payment_method_gateway.go
@@ -15,6 +15,7 @@ type PaymentMethodRequest struct {
 	Token              string                       `xml:"token,omitempty"`
 	PaymentMethodNonce string                       `xml:"payment-method-nonce,omitempty"`
 	Options            *PaymentMethodRequestOptions `xml:"options,omitempty"`
+	BillingAddress     *AddressRequest              `xml:"billing-address,omitempty"`
 }
 
 type PaymentMethodRequestOptions struct {
--- a/payment_method_integration_test.go
+++ b/payment_method_integration_test.go
@@ -25,9 +25,24 @@ func TestPaymentMethod(t *testing.T) {
 	g := testGateway.PaymentMethod()
 
 	// Create using credit card
+	addr := &AddressRequest{
+		FirstName:          "Robert",
+		LastName:           "Smith",
+		Company:            "The Cure",
+		StreetAddress:      "39 Acacia Avenue",
+		ExtendedAddress:    "SAV Studios",
+		Locality:           "North End",
+		Region:             "London",
+		PostalCode:         "SW1A 0AA",
+		CountryCodeAlpha2:  "GB",
+		CountryCodeAlpha3:  "GBR",
+		CountryCodeNumeric: "826",
+		CountryName:        "United Kingdom",
+	}
 	paymentMethod, err := g.Create(ctx, &PaymentMethodRequest{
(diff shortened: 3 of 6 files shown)


payment_method_gateway.go
```diff
@@ -15,6 +15,7 @@ type PaymentMethodRequest struct {
 	Token              string                       `xml:"token,omitempty"`
 	PaymentMethodNonce string                       `xml:"payment-method-nonce,omitempty"`
 	Options            *PaymentMethodRequestOptions `xml:"options,omitempty"`
+	BillingAddress     *Address                     `xml:"billing-address,omitempty"`
```
leighmcculloch: Can you change `Address` to `AddressRequest`? The latter is used in requests from the client to the API and the former is used in responses from the API to the client.

payment_method_integration_test.go
```diff
@@ -28,6 +28,9 @@ func TestPaymentMethod(t *testing.T) {
 	paymentMethod, err := g.Create(ctx, &PaymentMethodRequest{
 		CustomerId:         cust.Id,
 		PaymentMethodNonce: FakeNonceTransactableVisa,
+		BillingAddress: &Address{
+			CountryCodeAlpha2: "US",
```
leighmcculloch: Could you add all fields in the `AddressRequest` to this test to ensure they're all accepted by the API in this endpoint?

leighmcculloch (changes requested):
Thanks for submitting this PR @lindroth and also for adding tests. There are a couple things we need to change to merge it, see below for comments in line. Once they're changed we can merge it and get it in a release!

address.go
```diff
@@ -26,19 +26,18 @@ type Address struct {
 }
 
 type AddressRequest struct {
-	XMLName            xml.Name `xml:"address"`
```
lindroth, line 29: Had to remove `XMLName` as I got 
`payment_method_integration_test.go:48: xml: name "billing-address" in tag of braintree.PaymentMethodRequest.BillingAddress conflicts with name "address" in *braintree.AddressRequest.XMLName`
This should be fine. But I have some problems running all integrations test against my sandbox. @leighmcculloch can you verify this? Otherwise I would need to create a new type `BillingAddressRequest` with another `XMLName`

payment_method_integration_test.go
lindroth: Done

payment_method_gateway.go
lindroth: Fixed, but lead to some problems described below.

lindroth:
Ok, see that the tests fails on Travis. I think adding a new BillingAddress type is the way to go then.
What do you think @leighmcculloch?

address.go
leighmcculloch: I'll verify this.

leighmcculloch: So removing this makes this work for cases where `AddressRequest` is used as a sub-element where it's top-level XMLName comes from the parent, but `AddressRequest` is also used as a root XML node in the `AddressGateway` and this causes it to serialize with the root tag name `AddressRequest` instead of `address` which is not accepted by the API.

leighmcculloch: I think we can make `AddressRequest` work for both if we keep the `XMLName` field and set it's `Local` field to be the name we need in the `AddressGateway` case. e.g.:

```patch
From caa5b7554902b05c347d8fb1f24effc3c3374625 Mon Sep 17 00:00:00 2001
From: Leigh McCulloch <leigh@mcchouse.com>
Date: Wed, 17 Apr 2019 05:26:15 +0000
Subject: Fix use of AddressRequest in AddressGateway

---
 address.go         | 1 +
 address_gateway.go | 2 ++
 2 files changed, 3 insertions(+)

diff --git a/address.go b/address.go
index 9e9d3ee..6f2060d 100644
--- a/address.go
+++ b/address.go
@@ -26,6 +26,7 @@ type Address struct {
 }
 
 type AddressRequest struct {
+       XMLName            xml.Name
        FirstName          string `xml:"first-name,omitempty"`
        LastName           string `xml:"last-name,omitempty"`
        Company            string `xml:"company,omitempty"`
diff --git a/address_gateway.go b/address_gateway.go
index e0b7f73..8db254b 100644
--- a/address_gateway.go
+++ b/address_gateway.go
@@ -10,6 +10,7 @@ type AddressGateway struct {
 
 // Create creates a new address for the specified customer id.
 func (g *AddressGateway) Create(ctx context.Context, customerID string, a *AddressRequest) (*Address, error) {
+       a.XMLName.Local = "address"
        resp, err := g.execute(ctx, "POST", "customers/"+customerID+"/addresses", &a)
        if err != nil {
                return nil, err
@@ -36,6 +37,7 @@ func (g *AddressGateway) Delete(ctx context.Context, customerId, addrId string)
 
 // Update updates an address for the address id and customer id.
 func (g *AddressGateway) Update(ctx context.Context, customerID, addrID string, a *AddressRequest) (*Address, error) {
+       a.XMLName.Local = "address"
        resp, err := g.execute(ctx, "PUT", "customers/"+customerID+"/addresses/"+addrID, a)
        if err != nil {
                return nil, err
-- 
2.20.1
```

lindroth: Tried this. But the `Test_grpcServer_GetTransaction` test fails with:
`expected tx2.RiskData not to be empty`
 
@leighmcculloch, do you have any idea why this happens?

jvalecillos:
I think we are repeating ourselves for no reason. Consider composition by using embedded structs. For example: 

```go

// baseAddress represents the base struct for addresses
type baseAddress struct {
	FirstName          string `xml:"first-name,omitempty"`
	LastName           string `xml:"last-name,omitempty"`
	Company            string `xml:"company,omitempty"`
	StreetAddress      string `xml:"street-address,omitempty"`
	ExtendedAddress    string `xml:"extended-address,omitempty"`
	Locality           string `xml:"locality,omitempty"`
	Region             string `xml:"region,omitempty"`
	PostalCode         string `xml:"postal-code,omitempty"`
	CountryCodeAlpha2  string `xml:"country-code-alpha2,omitempty"`
	CountryCodeAlpha3  string `xml:"country-code-alpha3,omitempty"`
	CountryCodeNumeric string `xml:"country-code-numeric,omitempty"`
	CountryName        string `xml:"country-name,omitempty"`
}

// Address represents the address used on transaction requests/responses
type Address struct {
	XMLName    xml.Name
	Id         string `xml:"id,omitempty"`
	CustomerId string `xml:"customer-id,omitempty"`
	baseAddress
	CreatedAt *time.Time `xml:"created-at,omitempty"`
	UpdatedAt *time.Time `xml:"updated-at,omitempty"`
}

// AddressRequest represents the address used for creating payment methods.
type AddressRequest struct {
	XMLName xml.Name
	baseAddress
}
```

This way you don't have to repeat the fields and tags over and over.

lindroth:
@jvalecillos, I agree. But should that be part of this pull request, or in it's own pull request?

jvalecillos:
@lindroth Checking the code in master, I just noticed that my suggestion could actually be part of a different improvement including also other structs.

leighmcculloch:
@lindroth I just wanted to let you know [I have left Braintree](https://twitter.com/___leigh___/status/1146485107881996288) and I'm unable to continue contributing to this library.

